PR #14, opened 2022-04-25
fixing calling container cc accessibility issue with end call button

sarojkpr:
Fixing color contrast bug in calling endcall button, replacing white to hex


abhinab22mohanty (approved):
Would be great if you could add a screenshot of the color analyzer tool ratio for the colors updated

PR #17, opened 2022-04-28
Fixed LiveChatWidget import since its not default exported and Security link at the bottom

sudhcha:
The LiveChatWidget is not default exported from omnichannel-chat-widget. As a result, changed the reference code to use named import.


charliewang95 (approved):

charliewang95 (commented):
Line 70 is missing a "const". Can you please add it along with this PR? Thanks

sarojkpr (approved):

xTEddie (approved):

PR #45, opened 2022-06-08
package.json update for allowing to fetch package version number

sarojkpr:
1. Allowing omnichannel-chat-component and omnichannel-chat-widget packages to provide version info from respective packages.json.
2. Making telemetry config mandatory to avoid customers to use default version number.


chat-widget/src/common/telemetry/interfaces/ITelemetryConfig.ts
```diff
@@ -6,40 +6,49 @@ export interface ITelemetryConfig {
...
     /**
     * custom loggers list
     */
     telemetryLoggers?: IChatSDKLogger[];
+
     /**
     * Omnichannel chat widget version
     */
-    chatWidgetVersion?: string;
+    chatWidgetVersion: string;
```
charliewang95, line 43: Why are these not optional any more?

charliewang95 (approved):

PR #55, opened 2022-06-29
Adding font-family to preChatSurvey pane custom Button properties

abhinab22mohanty:
image


chat-components/src/components/prechatsurveypane/common/defaultProps/defaultStyles/defaultPreChatSurveyPaneButtonStyles.ts
```diff
@@ -2,5 +2,6 @@ import { IPreChatSurveyPaneButtonStyles } from "../../../interfaces/IPreChatSurv
 
 export const defaultPreChatSurveyPaneButtonStyles: IPreChatSurveyPaneButtonStyles = {
     backgroundColor: "rgb(49, 95, 162)",
-    color: "#FFFFFF"
+    color: "#FFFFFF",
+    fontFamily: "sans-serif"
```
marghayk, line 6: It looks like we use "Segoe UI, Arial, sans-serif" in most parts. Do we want to change to the same font family?

PR #103, opened 2022-08-25
Users/sarojkpr/broadcast service  widget isolation

sarojkpr:
Updating the Broadcast Service for multi widget/multi-tab scenario for listener isolation. This is needed to avoid widget2 also listening to SDK methods from same Broadcast channel.

### DIFF
--- a/chat-components/src/services/BroadcastService.ts
+++ b/chat-components/src/services/BroadcastService.ts
@@ -2,16 +2,40 @@ import { Subject } from "rxjs";
 import { ICustomEvent } from "../interfaces/ICustomEvent";
 import { filter } from "rxjs/operators";
 import { BroadcastChannel } from "broadcast-channel";
-import { BROADCAST_CHANNEL_NAME } from "../common/Constants";
 
 const newMessage = new Subject<ICustomEvent>();
 
-const pubChannel = new BroadcastChannel(BROADCAST_CHANNEL_NAME);
-const subChannel = new BroadcastChannel(BROADCAST_CHANNEL_NAME);
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const broadcastServicePubList: Record<string, any> = {};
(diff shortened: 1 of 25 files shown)


chat-components/src/services/BroadcastService.ts
```diff
@@ -2,16 +2,40 @@ import { Subject } from "rxjs";
 import { ICustomEvent } from "../interfaces/ICustomEvent";
 import { filter } from "rxjs/operators";
 import { BroadcastChannel } from "broadcast-channel";
-import { BROADCAST_CHANNEL_NAME } from "../common/Constants";
```
charliewang95, line 5: If not used anymore, can we remove from Constants?

charliewang95 (approved):

chat-components/src/services/BroadcastService.ts
```diff
@@ -2,16 +2,40 @@ import { Subject } from "rxjs";
...
+const broadcastServiceSubList: Record<string, any> = {};
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-subChannel.onmessage = (message: any) => {
-    newMessage.next(message);
+let pubChannel: any;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let subChannel: any;
+
+export const BroadcastServiceInitialize = (channelName: string) => {
```
charliewang95, line 18: Please check if all lcw events are still being raised from script layer after this change

charliewang95 (approved):

chat-components/src/services/BroadcastService.ts
sarojkpr: removed this

chat-components/src/services/BroadcastService.ts
sarojkpr: okay, will validate them before raising widget PR

PR #106, opened 2022-08-29
Users/sarojkpr/broadcast service update for multi widget

sarojkpr:
1. Adding widget specific broadcast service changes for SDK methods

image

### DIFF
--- a/chat-widget/src/common/Constants.ts
+++ b/chat-widget/src/common/Constants.ts
@@ -119,6 +119,7 @@ export class Constants {
 
     public static readonly ChatWidgetStateChangedPrefix = "ChatWidgetStateChanged";
     public static readonly PostChatLoadingDurationInMs = 2000;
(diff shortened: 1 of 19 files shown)


chat-widget/src/common/Constants.ts
```diff
@@ -119,6 +119,7 @@ export class Constants {
 
     public static readonly ChatWidgetStateChangedPrefix = "ChatWidgetStateChanged";
     public static readonly PostChatLoadingDurationInMs = 2000;
+    public static readonly BrowserUnloadConfirmationMessage = "Do you want to leave chat?";
```
charliewang95, line 122: Is this customizable? We need localization on this text

sarojkpr: Browsers have their default prompts and don't accepts customized string. This is just a dummy string for browsers to prompt the confirmation.

sarojkpr: image

charliewang95 (approved):

PR #107, opened 2022-08-31
Added custom context for popout & throw an exception incase of auth chat

dishakhattri:
Exception :
image

image

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/startChat.ts
+++ b/chat-widget/src/components/livechatwidget/common/startChat.ts
@@ -23,10 +23,13 @@ import { ActivityStreamHandler } from "./ActivityStreamHandler";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 let optionalParams: any = {};
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+let widgetInstanceId: any | "";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const prepareStartChat = async (props: ILiveChatWidgetProps, chatSDK: any, state: ILiveChatWidgetContext, dispatch: Dispatch<ILiveChatWidgetAction>, setAdapter: any) => {
     optionalParams = {}; //Resetting to ensure no stale values
+    widgetInstanceId = props?.controlProps?.widgetInstanceId;
 
     // Can connect to existing chat session
     if (await canConnectToExistingChat(props, chatSDK, state, dispatch, setAdapter)) {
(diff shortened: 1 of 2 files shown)


chat-widget/src/components/livechatwidget/common/startChat.ts
```diff
@@ -219,11 +217,21 @@ const canConnectToExistingChat = async (props: ILiveChatWidgetProps, chatSDK: an
...
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const setCustomContextParams = (props: ILiveChatWidgetProps, chatSDK: any) => {
+const setCustomContextParams = (chatSDK: any) => {
     // Add custom context only for unauthenticated chat
     const persistedState = getStateFromCache(chatSDK?.omnichannelConfig?.orgId, chatSDK?.omnichannelConfig?.widgetId, props?.controlProps?.widgetInstanceId ?? "");
 
-    if (!props.chatConfig?.LiveChatConfigAuthSettings && !isUndefinedOrEmpty(persistedState?.domainStates?.customContext)) {
+    if (!isUndefinedOrEmpty(persistedState?.domainStates?.customContext)) {
+        if(persistedState?.domainStates.liveChatConfig?.LiveChatConfigAuthSettings) {
+            const errorMessage = "Use of custom context with authenticated chat is deprecated. The chat would not go through.";
```
elopezanaya, line 226: why not use a constant value as the other values for the error object?

chat-widget/src/components/livechatwidget/common/startChat.ts
```diff
@@ -219,11 +217,21 @@ const canConnectToExistingChat = async (props: ILiveChatWidgetProps, chatSDK: an
 };
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const setCustomContextParams = (props: ILiveChatWidgetProps, chatSDK: any) => {
+const setCustomContextParams = (chatSDK: any) => {
     // Add custom context only for unauthenticated chat
     const persistedState = getStateFromCache(chatSDK?.omnichannelConfig?.orgId, chatSDK?.omnichannelConfig?.widgetId, props?.controlProps?.widgetInstanceId ?? "");
```
elopezanaya, line 222: if there any point to get persistedState here, when we may not use it?, I think a better place is after LOC 224,

elopezanaya (changes requested):
couple of comments, and validations are failing

chat-widget/src/components/livechatwidget/common/startChat.ts
dishakhattri: As talked over call, will fix as per another task for all hardcoded messges.

dishakhattri (commented):
Resolved comments

elopezanaya (approved):

PR #112, opened 2022-09-06
2904088_fix logic for markdown

elopezanaya:
# Description

it started as checking for the issue when using tags and the org doesnt have enabled markdown, the expected result is that the markdown is in action if presented the data tag.

# Solution 

- fix for wrong logic around flags disableMarkdownMessageFormatting (wrong flow)
-  for for creation of markdown object, which was broken, it was never creating a healthy object, so therefore render was not happening 
-  this fix an issue with links and erroneous render of markdown
-  support bold markdown for links ex -> **[hi](www.hi.com)**

# EVIDENCE

### org markdown disabled , no data tag present

image

### org markdown disabled , data tag present

image

### org markdown disabled , no data tag, custom config present

image

### org markdown disabled ,  data tag, custom config present to disable format enable

image

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/createMarkdown.ts
+++ b/chat-widget/src/components/livechatwidget/common/createMarkdown.ts
@@ -7,20 +7,25 @@ import { Constants } from "../../../common/Constants";
 export const createMarkdown = (disableMarkdownMessageFormatting: boolean, disableNewLineMarkdownSupport: boolean) => {
     let markdown: MarkdownIt;
 
-    if (!disableMarkdownMessageFormatting) {
+    if (disableMarkdownMessageFormatting) {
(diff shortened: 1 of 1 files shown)


chat-widget/src/components/livechatwidget/common/createMarkdown.ts
```diff
@@ -7,20 +7,25 @@ import { Constants } from "../../../common/Constants";
...
-            Constants.Default,
+            Constants.Zero,
             {
                 html: true,
                 linkify: true,
                 breaks: (!disableNewLineMarkdownSupport)
             }
         );
+
+        markdown.enable([
```
charliewang95, line 20: Why is this still needed if we already have line 15?

PR #113, opened 2022-09-07
Setting auth token before getting reconnectId

marghayk:
Setting auth token before getting reconnectId and removing it afterwards

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
+++ b/chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
@@ -16,30 +16,31 @@ import { TelemetryHelper } from "../../../common/telemetry/TelemetryHelper";
 import { handleAuthentication, removeAuthTokenProvider } from "./authHelper";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const getChatReconnectContext = async (chatSDK: any, chatConfig: ChatConfig | undefined, getAuthToken: ((authClientFunction?: string) => Promise<string | null>) | undefined, reconnectId?: string) => {
-    try {
-        if (reconnectId) {
-            const chatReconnectOptionalParams: IReconnectChatOptionalParams = {
-                reconnectId: reconnectId
-            };
-            return await chatSDK?.getChatReconnectContext(chatReconnectOptionalParams);
-        } else {
-            // set auth token to chat sdk to get reconnectId for auth chat
-            await handleAuthentication(chatSDK, chatConfig, getAuthToken);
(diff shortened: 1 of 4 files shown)


chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
```diff
@@ -13,25 +13,34 @@ import { IReconnectChatContext } from "../../reconnectchatpanestateful/interface
...
+                const reconnectChatContext = await chatSDK?.getChatReconnectContext();
+                // remove auth token after reconnectId is fetched
+                // this will be reset later at start chat
+                removeAuthTokenProvider(chatSDK);
+                return reconnectChatContext;
             }
-        });
+        } catch (ex) {
+            TelemetryHelper.logSDKEvent(LogLevel.ERROR, {
+                Event: TelemetryEvent.GetChatReconnectContextSDKCallFailed,
```
sarojkpr, line 38: Please ensure this is published to as lcw:error when adding script code

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
```diff
@@ -13,25 +13,34 @@ import { IReconnectChatContext } from "../../reconnectchatpanestateful/interface
...
+                    reconnectId: reconnectId
+                };
+                return await chatSDK?.getChatReconnectContext(chatReconnectOptionalParams);
+            } else {
+                // set auth token to chat sdk to get reconnectId for auth chat
+                await handleAuthentication(chatSDK, chatConfig, getAuthToken);
+                const reconnectChatContext = await chatSDK?.getChatReconnectContext();
+                // remove auth token after reconnectId is fetched
+                // this will be reset later at start chat
+                removeAuthTokenProvider(chatSDK);
```
sarojkpr, line 33: Seems we're moving the token, does it mean we'll get a new token for authentication? If yes, is reconnect working in the token changed for `startChat`?

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
marghayk: will do

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
marghayk: The token will be in a global variable in script level, so it will still be the same token. But if it is expired, we will get a new token. My understanding is that token is not relevant for reconnecting, and that it reconnects to previous chat based on reconnectId

sarojkpr (approved):

PR #128, opened 2022-09-28
Auth chat - connect to same chat after refresh

marghayk:
Currently when there is auth chat and we refresh the page or start in another tab, it doesn't connect to the same chat, but starts a new chat instead. With this fix it will connect to the same chat.


chat-widget/src/components/livechatwidget/common/startChat.ts
```diff
@@ -81,6 +81,15 @@ const initStartChat = async (chatSDK: any, chatConfig: ChatConfig | undefined, g
     try {
         let isStartChatSuccessful = false;
```
sarojkpr, line 82: I understand we got little lazy, but let's move the initialize line below authentication :)

sarojkpr (approved):

chat-widget/src/components/livechatwidget/common/startChat.ts
marghayk: done

PR #131, opened 2022-09-30
Replace proactive chat close button

marghayk:
Currently header and proactive chat pane have different close button. Header uses fluent ui icon, but for proactive chat we provide the icon image.

With this change we replace proactive chat pane's close button with header's close button.

proactivechatpane


dishakhattri (commented):
Update the ChangeLog and Readme files for the props change.

dishakhattri (approved):
could you please add a screenshot to showcase it in working state?

PR #143, opened 2022-10-10
Users/hmargaryan/fix conversation state

marghayk:
1) Currently when refreshing auth chat, if there is no auth token, it keeps the active state, even though start chat isn't called. Changing the state to show the chat button.

2) Adding loading pane, while it is getting auth token.

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/startChat.ts
+++ b/chat-widget/src/components/livechatwidget/common/startChat.ts
@@ -81,9 +81,11 @@ const initStartChat = async (chatSDK: any, chatConfig: ChatConfig | undefined, g
     try {
         const authClientFunction = getAuthClientFunction(chatConfig);
         if (getAuthToken && authClientFunction) {
+            dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.Loading });
             // set auth token to chat sdk before start chat
             const authSuccess = await handleAuthentication(chatSDK, chatConfig, getAuthToken);
             if (!authSuccess) {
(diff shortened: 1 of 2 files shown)


chat-widget/src/components/livechatwidget/common/startChat.ts
```diff
@@ -81,9 +81,11 @@ const initStartChat = async (chatSDK: any, chatConfig: ChatConfig | undefined, g
     try {
         const authClientFunction = getAuthClientFunction(chatConfig);
         if (getAuthToken && authClientFunction) {
+            dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.Loading });
             // set auth token to chat sdk before start chat
             const authSuccess = await handleAuthentication(chatSDK, chatConfig, getAuthToken);
             if (!authSuccess) {
+                dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.Closed });
```
sarojkpr, line 88: If there is a prob is auth token, and the state is set to close, I think it might show the Start Chat button. For customer, it would give an experience that Start chat button wasn't clicked, so customer will keep on clicking on it. Can you verify this behavior in embed mode? How about using `inactive` state or adding a new state?

marghayk: yep it is showing chat button. How about we use loading pane instead?

elopezanaya: can we include some screen shots how it looks in the scenario of failure ?

PR #147, opened 2022-10-18
Added Support for Download Transcript on PostChat Screen [V1 Parity] and fixed multiple scrollbar for PostChat

abhinab22mohanty:
image

Fixed the following bugs
1. [LCW][V2] PostChat Survey Feature Parity: Download transcript and footer view
2. [LCW][V2] Multiple scrollbars in post chat survey


chat-widget/src/components/headerstateful/HeaderStateful.tsx
```diff
@@ -34,7 +34,7 @@ export const HeaderStateful = (props: IHeaderStatefulParams) => {
         },
         onCloseClick: async () => {
             TelemetryHelper.logActionEvent(LogLevel.INFO, { Event: TelemetryEvent.HeaderCloseButtonClicked, Description: "Header Close button clicked." });
-            if (conversationState.current === ConversationState.Active) {
+            if (conversationState.current === ConversationState.Active || (conversationState.current === ConversationState.Postchat) && !isDownloadTranscriptDisabled && !isEmailTranscriptDisabled) {
```
sarojkpr, line 37: `!isDownloadTranscriptDisabled && !isEmailTranscriptDisabled`  maynot be combined as both of them are configurable individually.

chat-widget/src/components/headerstateful/interfaces/IHeaderStatefulParams.ts
```diff
@@ -13,6 +13,18 @@ export interface IHeaderStatefulParams {
      */
     outOfOfficeHeaderProps?: IHeaderProps;
 
+   /**
+    * isEmailTranscriptDisabled: Props to check if Email Transcript is Disabled
+    * These props are used for maintaining consistent behaviour between header and footer
```
sarojkpr, line 18: we shouldn't bring footer props to header; we need figure out a better design

sarojkpr (changes requested):
please check again if can we use `chatconfig` props

chat-widget/src/components/headerstateful/interfaces/IHeaderStatefulParams.ts
abhinab22mohanty: How about we have a separate boolean prop for header like:

```
    // Header prop to check if download transcript is enabled
    const shouldShowDownloadChatCopy = !(emailTranscriptDisabled && downloadTranscriptDisabled);
```

chat-widget/src/components/headerstateful/HeaderStateful.tsx
abhinab22mohanty: Correct Thanks for the catch we would need them in either case
`!emailTranscriptDisabled || !downloadTranscriptDisabled;`

abhinab22mohanty: We have to check for both because in either case we have "Get copy of chat" confirmation pane in V1

chat-widget/src/components/headerstateful/interfaces/IHeaderStatefulParams.ts
sarojkpr: Isn't confirmation pane capable to handle this message based on footer logic? All we had to do show the footer.

chat-widget/src/components/headerstateful/HeaderStateful.tsx
sarojkpr: lets keep the same message as it appears before post chat screen.

charliewang95 (approved):

abhinab22mohanty:
> please check again if can we use `chatconfig` props

sarojkpr (approved):

PR #153, opened 2022-10-21
Scrollbar customization capability

sarojkpr:
- Adding scrollbar customization capability
- Removed `tabIndex` from `div` in prechat survey for accessibility (Abhinab fix)

**Screenshots:**
Default:
image

Customized:
image

### DIFF
--- a/chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
+++ b/chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
@@ -63,6 +63,8 @@ import useChatSDKStore from "../../../hooks/useChatSDKStore";
 import { ActivityStreamHandler } from "../common/ActivityStreamHandler";
 import { registerBroadcastServiceForLocalStorage } from "../../../common/storage/default/defaultCacheManager";
 import { defaultClientDataStoreProvider } from "../../../common/storage/default/defaultClientDataStoreProvider";
+import { IScrollBarProps } from "../interfaces/IScrollBarProps";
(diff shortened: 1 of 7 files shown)


chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
```diff
@@ -426,44 +431,64 @@ export const LiveChatWidgetStateful = (props: ILiveChatWidgetProps) => {
...
     return (
-        <Composer
-            {...webChatProps}
-            styleOptions={webChatStyles}
-            directLine={props.webChatContainerProps?.directLine ?? adapter ?? defaultWebChatContainerStatefulProps.directLine}>
-            <Stack
-                id={widgetElementId}
-                styles={generalStyles}
-                className={props.styleProps?.className}>
+        <>
```
abhinab22mohanty, line 434: Can we check if there is any option to customize the default scrollbar with fluent UI instead of adding style tag here?

abhinab22mohanty (approved):

chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
sarojkpr: we want to allow customization for scrollbar instead of using default one.

abhinab22mohanty: I meant is there a way to use Fluent UI for scrollbar customization
Seems like tit is not possible https://github.com/microsoft/fluentui/issues/8302

PR #157, opened 2022-11-02
Refactoring input validation pane

marghayk:
1) Used useCallback hook
2) Replaced keydown with keyup for input field, as keydown causes an issue later in the stitching layer by creating an extra keyup or keydown event on focused element. Here is what happens in the stitching layer:
Press Enter on the input field -> it closes email transcript pane and puts the focus on email transcript footer action -> it creates an extra keydown or keyup event on that focused element, which causes email transcript pane to show again.

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -69,4 +69,4 @@ All notable changes to this project will be documented in this file.## [Unreleas
 - Fixed OutOfOffice Chat Button Icon appearance
 - Adding support for customizing widget scroll bar
 - Clearing live chat context for raise error event
-- Refactoring input validation pane
+- Refactoring input validation pane and replacing keyDown event with keyUp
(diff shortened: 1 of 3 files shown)


CHANGE_LOG.md
```diff
@@ -68,4 +68,5 @@ All notable changes to this project will be documented in this file.## [Unreleas
 - Removed isReconnectEnabled from props
 - Fixed OutOfOffice Chat Button Icon appearance
 - Adding support for customizing widget scroll bar
-- Clearing live chat context for raise error event
\ No newline at end of file
+- Clearing live chat context for raise error event
+- Refactoring input validation pane
```
abhinab22mohanty, line 72: Please add the Key up and Key down issue fix in ChangeLog

abhinab22mohanty (approved):

CHANGE_LOG.md
marghayk: Added

PR #161, opened 2022-11-09
Fix download and email transcript after post chat survey is sent / rendered

kevinlin98z:
**Issue:**
Download and email transcript does not work after Post Chat Survey is displayed. Investigations traced issue to endChat SDK call happening after PCS flow is triggered, which resets the chatToken (which has chatId) and requestId necessary for the download and email transcript API calls.
image
Here we see the chatId as part of the API request is undefined, which gives 500 internal server error.
https://unq2ef66d93e33c443aadbbbf19627d3-crm10.oc.crmlivetie.com/livechatconnector/v2/getchattranscripts/**undefined**/6ffd4907-11ac-4af6-8fad-c0e2cae5db41?channelId=lcw

**Solution:**
Save the chatToken and requestId before the endChat SDK call is made, and re-assign those values to the chatSDK object such that the download and email transcript API calls can be made successfully.

**Testing:**
Tested download and email transcript for link mode PCS, embed mode PCS, and no PCS, with C1 and C2 ending for the aforementioned scenarios. Here's some screenshots for link mode PCS testing:
image
image
image
image

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/endChat.ts
+++ b/chat-widget/src/components/livechatwidget/common/endChat.ts
@@ -58,6 +58,7 @@ const prepareEndChat = async (props: ILiveChatWidgetProps, chatSDK: any, setAdap
         const chatSession = await chatSDK?.getCurrentLiveChatContext();
         await endChat(props, chatSDK, setAdapter, setWebChatStyles, dispatch, adapter, skipEndChatSDK, skipCloseChat, false);
         if (chatSession) {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
             (chatSDK as any).chatToken = chatSession.chatToken ?? {};
             chatSDK.requestId = chatSession.requestId;
         }
--- a/chat-widget/src/components/livechatwidget/common/endChat.ts
+++ b/chat-widget/src/components/livechatwidget/common/endChat.ts
@@ -55,8 +55,13 @@ const prepareEndChat = async (props: ILiveChatWidgetProps, chatSDK: any, setAdap
     if (isPostChatEnabled === "true" && conversationDetails?.canRenderPostChat === Constants.truePascal) {
         const skipEndChatSDK = false;
         const skipCloseChat = true;
+        const chatSession = await chatSDK?.getCurrentLiveChatContext();
         await endChat(props, chatSDK, setAdapter, setWebChatStyles, dispatch, adapter, skipEndChatSDK, skipCloseChat, false);
-
+        if (chatSession) {
+            (chatSDK as any).chatToken = chatSession.chatToken ?? {};
+            chatSDK.requestId = chatSession.requestId;
+        }
+        
         if (postChatSurveyMode === PostChatSurveyMode.Embed) {
             dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.PostchatLoading });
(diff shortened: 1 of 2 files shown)


chat-widget/src/components/livechatwidget/common/endChat.ts
```diff
@@ -55,8 +55,13 @@ const prepareEndChat = async (props: ILiveChatWidgetProps, chatSDK: any, setAdap
     if (isPostChatEnabled === "true" && conversationDetails?.canRenderPostChat === Constants.truePascal) {
         const skipEndChatSDK = false;
         const skipCloseChat = true;
+        const chatSession = await chatSDK?.getCurrentLiveChatContext();
         await endChat(props, chatSDK, setAdapter, setWebChatStyles, dispatch, adapter, skipEndChatSDK, skipCloseChat, false);
-
+        if (chatSession) {
+            (chatSDK as any).chatToken = chatSession.chatToken ?? {};
```
abhinab22mohanty, line 61: **Minor**: Please disable eslint warning for any

chat-widget/src/components/livechatwidget/common/endChat.ts
```diff
@@ -55,8 +55,13 @@ const prepareEndChat = async (props: ILiveChatWidgetProps, chatSDK: any, setAdap
...
+            chatSDK.requestId = chatSession.requestId;
```
abhinab22mohanty, line 62: Don't we need to handle and clear the context and token after chat has ended or else it might cause issues with new chat instances.
In this case I think what we can do is clear the context and token in the close chat part

kevinlin98z: The clearing of the context and token is already handled in the endChat method within endChat.ts file. See https://github.com/microsoft/omnichannel-chat-widget/blob/f0b26a69d61a7c3be5d023d83ad7ae7e92b06b6a/chat-widget/src/components/livechatwidget/common/endChat.ts#L101

xTEddie: Since we're reusing the requestId & chatId of the conversation that was closed to download/email the transcript, I'd recommend to reset those values after the chat widget is fully closed (when download/email transcript won't be available anymore).

On `ChatSDK.endChat()`, it basically recreates a new `requestId` via `uuidv4()` & sets the `chatToken` to be `{}`. You can use these values on reset.

abhinab22mohanty: I agree we should be resetting the values ideally to avoid any side effect.
Please verify if the values are being reset after fully closing chat in this case

chat-widget/src/components/livechatwidget/common/endChat.ts
kevinlin98z: done

chat-widget/src/components/livechatwidget/common/endChat.ts
kevinlin98z: We are already resetting the values in the endChat method of endChat.ts:
image
endChat flow will get called when C2 officially ends the chat (chat widget minimized and download/email transcript no longer available).

I verified that the values are in fact already getting reset through console logs added. And starting new chats do not use the previous chatToken/requestId.
image

abhinab22mohanty (approved):

xTEddie (approved):

PR #176, opened 2022-12-16
Added more customization to prechat survey pane

marghayk:
Added height, fontSize and paddingTop to prechat survey pane props. This will be used to fix element sizes for popout mode in mobile browsers


chat-components/src/components/prechatsurveypane/common/defaultProps/defaultStyles/defaultPreChatSurveyPaneButtonStyles.ts
```diff
@@ -3,5 +3,7 @@ import { IPreChatSurveyPaneButtonStyles } from "../../../interfaces/IPreChatSurv
 export const defaultPreChatSurveyPaneButtonStyles: IPreChatSurveyPaneButtonStyles = {
     backgroundColor: "rgb(49, 95, 162)",
     color: "#FFFFFF",
-    fontFamily: "Segoe UI, Arial, sans-serif"
+    fontFamily: "Segoe UI, Arial, sans-serif",
```
akagar-microsoft, line 6: Is this font-family referred on all the devices ? If so, we might need to think of font support for iOS devices
E.g., something like this
-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;

marghayk: This font family is used in other components as well. I will discuss this with the team

abhinab22mohanty (approved):

chat-components/src/components/prechatsurveypane/common/defaultProps/defaultStyles/defaultPreChatSurveyPaneButtonStyles.ts
marghayk: Added a User Story for this: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3114058

PR #197, opened 2023-01-25
pre chat survey pane narrator issue

v-hghildiyal:
[Bug 3047419](https://dev.azure.com/dynamicscrm/OneCRM/_workitems/edit/3047419): A11y_Live Chat Widget - Version 2_Live Chat_conversation with pre-chat_Screen reader: NVDA/Narrator is announcing the same information, "Fields marked with * are mandatory", twice

Reason :- Message was appended again in button parent div aria-label , so it was reading again. As in attached screenshot.

Fix:- added condition in PreChatSurveyPaneStateful where aria-label is set to field so that it exclude div with class "ac-actionSet".
Now narrator reads only once.

image


charliewang95 (approved):

chat-widget/src/components/prechatsurveypanestateful/PreChatSurveyPaneStateful.tsx
```diff
@@ -117,7 +117,10 @@ export const PreChatSurveyPaneStateful = (props: IPreChatSurveyPaneStatefulParam
                 }
                 if (current && current.tagName.toLowerCase() == HtmlAttributeNames.div && current.childElementCount > 0) {
                     const input = current.children[0].children;
-                    if (input && input.length > 0 && input[0].className != HtmlAttributeNames.adaptiveCardToggleInputClassName) {
+                    if (input
+                        && input.length > 0
```
charliewang95, line 121: Can simplify to `input?.length`?

charliewang95 (commented):
Please modify CHANGE_LOG.md

elopezanaya (changes requested):
Fix should work for NVDA and Narrator

chat-widget/src/components/prechatsurveypanestateful/PreChatSurveyPaneStateful.tsx
v-hghildiyal: Applied as mentioned.

PR #209, opened 2023-02-08
Adding telemetry for history message received

sarojkpr:
Partners claimed that they never received messagereceived or historymessagereceived event at their end, and currently LCW doesn't log historymessagereceived event, so its hard to troubleshoot these issues.

Fix: we're adding historymessagereceived event too to LCW kusto.


chat-widget/src/plugins/newMessageEventHandler.ts
```diff
@@ -77,6 +77,12 @@ export const createOnNewAdapterActivityHandler = (chatId: string, userId: string
                         Description: "New message received",
                         Data: payload
                     });
+                } else {
+                    TelemetryHelper.logActionEvent(LogLevel.INFO, {
```
charliewang95, line 81: This is going to log all messages every 30 seconds. Please check if this will overflow out database

sarojkpr: Updated to have only one historymessage
image

elopezanaya (approved):

docs/Telemetry.md
```diff
@@ -214,7 +214,7 @@ Refer to the below table to understand different critical telemetry events raise
 | `GetAuthTokenFailed` | On getting auth token failed |
 | `ReceivedNullOrEmptyToken` | On receiving null or empty auth token |
 | `SystemMessageReceived` | On system message received |
-
+| `HistoryMessageReceived` | On history message received |
```
charliewang95, line 217: Let's make a note here that it's only raised once

charliewang95 (approved):

docs/Telemetry.md
sarojkpr: okay, Updating it in next PR.

PR #210, opened 2023-02-09
Fix bug 3138034 part 1

josuhern:
Fixing the following items of [bug 3138034](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3138034) as I agreed with Charlie 

1. Loading pane (landscape mode + popout mode)
2. Input validation pane (landscape mode + popout mode)

hiding Icon when screen size is less than 375px
image
image
image

adding scroll bar to input validation pane when screen size gets smaller
image
image
image
image

image


chat-components/src/components/loadingpane/interfaces/ILoadingPaneProps.ts
```diff
@@ -6,4 +6,6 @@ export interface ILoadingPaneProps {
     componentOverrides?: ILoadingPaneComponentOverrides;
     controlProps?: ILoadingPaneControlProps;
     styleProps?: ILoadingPaneStyleProps;
+    windowWidth?: number;
```
kevinlin98z, line 9: Are we using windowWidth anywhere?

PR #227, opened 2023-03-15
Adding an event listener in case the chat window is hidden for notifications

dishakhattri:
- Adding support in case the chat window is hidden

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -97,3 +97,4 @@ All notable changes to this project will be documented in this file.## [Unreleas
 - Fixed PostChat loading twice when Bot enabled
 - Adding `hideErrorUIPane` to `ILiveChatWidgetControlProps` to show/hide Error UI on start chat
 - Fix Start Chat icon disappearing due to subtitle length
+- Adding support in case the chat window is hidden
--- a/chat-widget/src/common/telemetry/TelemetryConstants.ts
+++ b/chat-widget/src/common/telemetry/TelemetryConstants.ts
@@ -49,7 +49,6 @@ export enum BroadcastEvent {
     RaiseErrorEvent = "RaiseErrorEvent",
     NetworkDisconnected = "NetworkDisconnected",
     NetworkReconnected = "NetworkReconnected",
-    initiatePopoutChat = "initiatePopoutChat",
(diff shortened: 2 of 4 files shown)


CHANGE_LOG.md
```diff
@@ -97,3 +97,4 @@ All notable changes to this project will be documented in this file.## [Unreleas
 - Fixed PostChat loading twice when Bot enabled
 - Adding `hideErrorUIPane` to `ILiveChatWidgetControlProps` to show/hide Error UI on start chat
 - Fix Start Chat icon disappearing due to subtitle length
+- Adding support in case the chat window is hidden
```
charliewang95, line 100: Please elaborate more on this note

docs/Telemetry.md
```diff
@@ -138,6 +138,9 @@ Refer to the below table to understand different broadcast events raised during
 | `InitiateEndChatOnBrowserUnload`      | End active chats on browser unload |
 | `ClosePopoutWindow`      | Event to close popout window  |
 | `RaiseErrorEvent`                 | On raising error events |
+| `NetworkDisconnected`                 | On network Disconnects |
+| `NetworkReconnected`                 | On network reconnects |
+| `HideChatVisibilityChangeEvent`                 | On hiding the chat window or tab |
```
charliewang95, line 143: This is not something chat-widget send, but receives. We don't need to log it here

charliewang95 (approved):

chat-widget/src/common/telemetry/TelemetryConstants.ts
```diff
@@ -52,6 +52,7 @@ export enum BroadcastEvent {
     SigninCardReceived = "SignInCardReceived",
     BotAuthConfigRequest = "BotAuthConfigRequest",
     BotAuthConfigResponse = "BotAuthConfigResponse",
+    HideChatVisibilityChangeEvent = "hideChatVisibilityChangeEvent"
```
marghayk, line 55: nit: Upper case for consistency?

marghayk (approved):

chat-widget/src/common/telemetry/TelemetryConstants.ts
marghayk: please disregard, it looks like incoming event is lower case too

PR #232, opened 2023-03-16
Fix error UI pane string override

kevinlin98z:


chat-widget/src/components/loadingpanestateful/LoadingPaneStateful.tsx
```diff
@@ -33,13 +33,13 @@ export const LoadingPaneStateful = (props: ILoadingPaneProps) => {
     };
 
     const errorUIControlProps: ILoadingPaneControlProps = {
+        ...props.controlProps,
         id: "oc-lcw-alert-pane",
         dir: state.domainStates.globalDir,
-        titleText: "Chat is failing to load.",
-        subtitleText: "Please Close the chat and try again.",
+        titleText: "We are unable to load chat at this time.",
```
charliewang95, line 39: "load chat" or "load the chat"

kevinlin98z: String has been confirmed to be the above by PM.

abhinab22mohanty (approved):

chat-widget/src/components/loadingpanestateful/LoadingPaneStateful.tsx
akagar-microsoft, line 39: Are we suppose to hard-code the strings ? I mean how does localization works in this case ?

kevinlin98z:
PR will soon be discarded.

PR #255, opened 2023-04-05
A11y ford

elopezanaya:
## 3226191 - Issue 2 – Elements must only use allowed ARIA attributes -

Removing aria attributes and title for chat button, relying on text on display

image

## 3226201 : Tittle attributes used as labels for buttons
 remove title attributes for :  
- minimize button
- close button

image

- download chat
- email
- sound

image

image

image

upload file and send button are not in our control.


chat-components/src/components/chatbutton/common/defaultProps/defaultChatButtonControlProps.ts
```diff
@@ -5,10 +5,7 @@ export const defaultChatButtonControlProps: IChatButtonControlProps = {
     id: Ids.DefaultChatButtonId,
     dir: "ltr",
     role: "button",
-    ariaLabel: AriaLabels.LetsChatWeAreOnline,
```
charliewang95, line 8: Why are these removed?

chat-components/src/components/common/commandbutton/CommandButton.tsx
```diff
@@ -66,6 +66,7 @@ function CommandButton(props: ICommandButtonProps) {
                     disabled={props.disabled}
                     className={props.className}
                     styles={buttonStyles}
+                    title={props.hideButtonTitle ? undefined : props.ariaLabel}
```
charliewang95, line 69: Why are we using title not ariaLabel?

chat-components/src/components/chatbutton/common/defaultProps/defaultChatButtonControlProps.ts
elopezanaya: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3226191 , 

the ask is to remove title and aria label

chat-components/src/components/common/commandbutton/CommandButton.tsx
elopezanaya: I can  add the aria label, but for what I see , the aria label is generally set at specific component level (Email, download, etc)

charliewang95 (approved):

xTEddie (approved):

PR #258, opened 2023-04-07
Prevent new chat creation failure after popout Proactive chat

dishakhattri:
Issue: Once starting the proactive chat, the new chat following it was not getting created.
Fix:  Currently, all the logic of starting a new chat only if there is no ongoing chat has been moved to "StartChat". Proactive chat workflow doesn't need to specifically check for the same. Thus, removing the temporary state maintained for Proactive chat.


sarojkpr (approved):
Please add some details to description what fixed the issue

CHANGE_LOG.md
```diff
@@ -61,4 +61,8 @@ All notable changes to this project will be documented in this file.
 
 - Added individual UI components like `Header`, `Footer`, and `ChatButton`, etc. with customizable `controlProps` and `styleProps`
 - Added `BroadcastService` to communicate with package consumer layer, and for telemetry purposes
-- Exposed `encodeComponentString` and `encodeComponentString`
\ No newline at end of file
+- Exposed `encodeComponentString` and `encodeComponentString`
```
charliewang95, line 64: Please add to the correct section chat-widget/unreleased

charliewang95 (approved):

PR #261, opened 2023-04-10
Update data masking logic to break loop on empty string rule

charliewang95:
[Bug 3298417](https://dev.azure.com/dynamicscrm/OneCRM/_workitems/edit/3298417): [V2] When a data masking rule matches empty string, it should not hang the chat

## Description
When a data masking rule matches empty strings, the while loop will go on forever

## Solution Proposed
Break the loop if the modified string is the same as the original string

## Acceptance criteria
Chat does not hang when trying to send a message with a data masking rule matching empty strings

## Test cases and evidence
https://lcwstorageaccountdev.blob.core.windows.net/charlwan/test/nam.html

### DIFF
--- a/chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/storemiddlewares/dataMaskingMiddleware.ts
+++ b/chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/storemiddlewares/dataMaskingMiddleware.ts
@@ -77,11 +77,6 @@ const applyDataMasking = (action: IWebChatAction, regexCollection: IDataMaskingR
                 });
             }
         }
-
-        // Exit if rule matched
-        if (isRuleMatched === true) {
-            break;
-        }
     }
 
     action.payload.text = text;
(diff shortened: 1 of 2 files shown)


chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/storemiddlewares/dataMaskingMiddleware.ts
```diff
@@ -37,7 +37,13 @@ const applyDataMasking = (action: IWebChatAction, regexCollection: IDataMaskingR
                 // eslint-disable-next-line no-cond-assign
                 while (match = regex.exec(text)) {
                     const replaceStr = match[0].replace(/./gi, maskedChar);
-                    text = text.replace(match[0], replaceStr);
+                    const modifiedText = text.replace(match[0], replaceStr);
```
akhiloct1510, line 40: Same comment as V1 

Just wondering if we need to add anything on Telemetry side unless it is not flooding the system.

Like based on some boolean flag and and outside the loop

charliewang95: For V2 is it hard to log it only once, since the input rules are not static and there's no good way to hold a global variable.

There's log if data masking rule DOES apply, so I think this will suffice

akhiloct1510 (approved):

kevinlin98z (approved):

PR #263, opened 2023-04-17
[Bug 3321448] - [LCW v2] custom close button should trim and show "..."

josuhern:
**Description**
This PR fixes the styling problem when data-custom-close-button-text is to long the size of the button should be that big
also fixes the problem on the missing tooltip on hover of the close button when it is a text

**Solution Proposed**
Adding the customizable style parameter for the button label, since we are using fluent ui the [default button](https://developer.microsoft.com/en-us/fluentui#/controls/web/button) label gets generated inside few <span> tags the way to style this not simple when using styleProps

**Acceptance criteria**
The component should have the capability to accept styleProps that affect the button label and should be able to trim the text just as LCW v1 does

**Test cases and evidence**
This is V1
image

previous V2
image

current V2 after change
image

https://josue.blob.core.windows.net/tasks/indexv2.html


chat-components/src/components/header/__screenshots__/stateless-components-header-azure-header-azure-header-chromium-snap.png
charliewang95, line 1: Do not modify existing test cases. Please revert the storybook changes

charliewang95: If needed create new screenshots, but I think we're good with script layer tests

chat-components/src/components/header/common/defaultProps/azureHeaderControlProps.ts
```diff
@@ -20,7 +20,7 @@ export const azureHeaderControlProps: IHeaderControlProps = {
     closeButtonProps: {
         id: "oc-lcw-header-closebutton",
         type: ButtonTypes.Text,
-        text: "End Chat"
+        text: "Close this Chat Widget"
```
charliewang95, line 23: Please revert

chat-components/src/components/header/Header.stories.tsx
```diff
@@ -62,6 +62,12 @@ const azureCustomHeaderProps: IHeaderProps = {
         },
         closeButtonHoverStyleProps: {
             color: "blue"
+        },
+        closeButtonLabelStyleProps: {
```
charliewang95, line 66: We can remove this from storybook too

chat-components/src/components/common/commandbutton/CommandButton.tsx
```diff
@@ -32,6 +32,10 @@ function CommandButton(props: ICommandButtonProps) {
                 ":focus .ms-Button-icon": {
                     // eslint-disable-next-line @typescript-eslint/no-explicit-any
                     ...((props?.focusStyles as any)?.icon as any)
+                },
+                ".ms-Button-label": {
```
charliewang95, line 36: Is this part necessary? This looks a bit hacky

CHANGE_LOG.md
```diff
@@ -67,6 +67,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Added customizable `styleProp` labelStyles to Close button component to style the button text
```
charliewang95, line 70: Add `### Added` section header

charliewang95 (approved):

PR #274, opened 2023-04-27
Exposing new prop to avoid automatically starting a new chat if already expired

marghayk:
1) Adding allowStartNewChatIfExpired to ILiveChatWidgetProps.

Do not start chat automatically when when allowStartNewChatIfExpired is false, but send a broadcast event instead. 

Below is the UI:

**Embed**:
allowStartNewChatIfExpiredEmbed

**hideStartChatButton**:
allowStartNewChatIfExpiredHideChatButton

**Popout**:
allowStartNewChatIfExpiredPopout

2) Fixing reconnect chat pane, when hideStartChatButton is true. Without this fix, if hideStartChatButton is true and we try to reconnect in a different browser, or InPrivate browser, reconnect pane doesn't show.

The following bootstrapper link includes the changes:
https://nisrastorageaccount.blob.core.windows.net/lcw2testing/dist/v2scripts/LiveChatBootstrapper.js

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -6,9 +6,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- allowStartNewChatIfExpired to ILiveChatWidgetProps
+
 ## Fixed
 
 - Fixed custom context not showing for popout chat
+- Fixed reconnect chat pane not showing, when hideStartChatButton is true
 
 ## [1.0.3] - 2023-4-24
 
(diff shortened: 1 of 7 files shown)


nareshme2016:
We should not do this kind of customer specific change. If time is concerns customer can extend reconnect time from admin settings.

sarojkpr:
> We should not do this kind of customer specific change. If time is concerns customer can extend reconnect time from admin settings.

@naresh: this change has nothing to do with reconnect time. With this change, we're only ensuring when this flag is passed in as `false`, not to start new chat automatically. This is to handle use cases when partners have requirements to have "one chat per case".

PR #287, opened 2023-05-17
Finalize downloadable typescript sample project

kevinlin98z:
Updated package versions, cleaned up documentation, verified quick start flow.

Typescript sample project can be built and started by running `yarn install`, `yarn build`, `yarn start` within typescript sample project root.

As per PM requirements:
1. `defaultProps` are clearly defined with option for user to quickly modify as necessary.
2. Sample set of customizations (to chat button, header, webchat, etc) are provided.

image
image
image

Note: default typescript project with no customizations will be created in future PRs.


xTEddie (approved):

xTEddie:
I see `defautlProps.ts` is huge. Is there any way that the file size could be refactored/reduced? As an user perspective, it may be hard to navigate through the file.

kevinlin98z:
> I see `defautlProps.ts` is huge. Is there any way that the file size could be refactored/reduced? As an user perspective, it may be hard to navigate through the file.

This was part of the discussed requirements from PM to share a visible list of all the defaultProps so users can see what types of properties can be modified easily. We can review again with PM/team to see if any other changes are needed.

charliewang95 (approved):

PR #290, opened 2023-05-23
Aria Logger Fix

xTEddie:
- Fix `AriaTelemetryLogger` to be initialized to pick the proper `collectorUI` on every logging event
- Update `environmentVersion` to be `prod` by default in `defaultInternalTelemetryData`

### DIFF
--- a/chat-widget/src/common/telemetry/defaultConfigs/defaultTelemetryInternalData.ts
+++ b/chat-widget/src/common/telemetry/defaultConfigs/defaultTelemetryInternalData.ts
@@ -1,7 +1,7 @@
 import { IInternalTelemetryData } from "../interfaces/IInternalTelemetryData";
 
 export const defaultInternalTelemetryData: IInternalTelemetryData = {
-    environmentVersion: "test",
+    environmentVersion: "prod",
     chatWidgetLocaleLCID: "1033",
     channelId: "lcw2.0",
 };
--- a/chat-widget/src/common/telemetry/loggers/ariaTelemetryLogger.ts
+++ b/chat-widget/src/common/telemetry/loggers/ariaTelemetryLogger.ts
@@ -8,14 +8,30 @@ import { Constants, AriaTelemetryConstants, EnvironmentVersion } from "../../Con
 import { IChatSDKLogger } from "../interfaces/IChatSDKLogger";
 import { TelemetryManager } from "../TelemetryManager";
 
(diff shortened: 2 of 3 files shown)


chat-widget/src/common/telemetry/defaultConfigs/defaultTelemetryInternalData.ts
```diff
@@ -1,7 +1,7 @@
 import { IInternalTelemetryData } from "../interfaces/IInternalTelemetryData";
 
 export const defaultInternalTelemetryData: IInternalTelemetryData = {
-    environmentVersion: "test",
+    environmentVersion: "prod",
```
elopezanaya, line 4: can we add this change to the change.log?, to communicate the switch from test to prod as default

akhiloct1510, line 4: Will this start pushing all the telemetry to prod instances irrespective of environment, like ocdailym telemetry flowing to prod

xTEddie: Telemetry is already pushed to prod by default. `environmentVersion` is a hardcoded value. The logger has a logic to ignore `collectorURI` if `environmentVersion` is not set to `prod`

chat-widget/src/common/telemetry/loggers/ariaTelemetryLogger.ts
```diff
@@ -8,14 +8,30 @@ import { Constants, AriaTelemetryConstants, EnvironmentVersion } from "../../Con
 import { IChatSDKLogger } from "../interfaces/IChatSDKLogger";
 import { TelemetryManager } from "../TelemetryManager";
 
+const AWTDefaultConfiguration = {
+    collectorUri: "https://browser.pipe.aria.microsoft.com/Collector/3.0/",
+    cacheMemorySizeLimitInNumberOfEvents: 10000,
+    disableCookiesUsage: false,
+    canSendStatEvent: (eventName: string) => { return true; }, // eslint-disable-line @typescript-eslint/no-unused-vars
```
kevinlin98z, line 15: What is this being used for?

xTEddie: I believe it's the callback to whether to send telemetry. It's the default configuration from Aria that was reused.

PR #333, opened 2023-07-27
[HOLD] Use hooks for startChat logic, fixing custom context disappearing issue

charliewang95:
This PR is the bare minimum to fix the custom context bug, and 90% of the logics are copied/pasted from static functions into hooks.

### DIFF
--- a/chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
+++ b/chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
@@ -85,6 +85,7 @@ import useChatContextStore from "../../../hooks/useChatContextStore";
 import useChatSDKStore from "../../../hooks/useChatSDKStore";
 import { defaultAdaptiveCardStyles } from "../../webchatcontainerstateful/common/defaultStyles/defaultAdaptiveCardStyles";
 import StartChatErrorPaneStateful from "../../startchaterrorpanestateful/StartChatErrorPaneStateful";
+import { StartChatFailureType } from "../../../contexts/common/StartChatFailureType";
 
 export const LiveChatWidgetStateful = (props: ILiveChatWidgetProps) => {
     const [state, dispatch]: [ILiveChatWidgetContext, Dispatch<ILiveChatWidgetAction>] = useChatContextStore();
@@ -203,6 +204,7 @@ export const LiveChatWidgetStateful = (props: ILiveChatWidgetProps) => {
 
     useEffect(() => {
         dispatch({ type: LiveChatWidgetActionType.SET_START_CHAT_FAILING, payload: false });
+        dispatch({ type: LiveChatWidgetActionType.SET_START_CHAT_FAILURE_TYPE, payload: StartChatFailureType.Generic });
         state.domainStates.confirmationPaneConfirmedOptionClicked = false;
         state.domainStates.confirmationState = ConfirmationState.NotSet;
         setupClientDataStore();
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - Enabled customizations for the start chat error pane by adding new `IStartChatErrorPaneProps` interface
+- Add specific error strings to start chat error pane for authentication related failure scenarios by including new texts to `IStartChatErrorPaneControlProps`
 
 ### Fixed
 
--- a/chat-widget/src/common/Constants.ts
+++ b/chat-widget/src/common/Constants.ts
@@ -303,6 +303,10 @@ export class StartChatErrorPaneConstants {
     public static readonly DefaultStartChatErrorPaneId = "oc-lcw-start-chat-error-pane";
     public static readonly DefaultStartChatErrorTitleText = "We are unable to load chat at this time.";
     public static readonly DefaultStartChatErrorSubtitleText = "Please try again later.";
+    public static readonly DefaultStartChatErrorUnauthorizedTitleText = "Chat authentication has failed.";
+    public static readonly DefaultStartChatErrorAuthSetupErrorTitleText = "Chat authentication has failed.";
+    public static readonly DefaultStartChatErrorUnauthorizedSubtitleText = "UNAUTHORIZED";
+    public static readonly DefaultStartChatErrorAuthSetupErrorSubtitleText = "AUTH SETUP ERROR";
 }
 
 export class AriaTelemetryConstants {
--- a/chat-widget/src/components/livechatwidget/common/startChatErrorHandler.spec.ts
+++ b/chat-widget/src/components/livechatwidget/common/startChatErrorHandler.spec.ts
@@ -36,7 +36,7 @@ describe("startChatErrorHandler unit test", () => {
                 Exception: `Widget load complete with error: Error: ${WidgetLoadCustomErrorString.AuthenticationFailedErrorString}`
             })
         }));
-        expect(dispatch).toHaveBeenCalledTimes(2);
+        expect(dispatch).toHaveBeenCalledTimes(3);
         expect(dispatch).toHaveBeenCalledWith(expect.objectContaining({
             type: LiveChatWidgetActionType.SET_CONVERSATION_STATE
         }));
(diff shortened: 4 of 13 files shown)


chat-widget/src/components/chatbuttonstateful/ChatButtonStateful.tsx
```diff
@@ -1,49 +1,53 @@
...
+        TelemetryHelper.logActionEvent(LogLevel.INFO, {
+            Event: TelemetryEvent.LCWChatButtonClicked
+        });
+        
+        if (state.appStates.isMinimized) {
+            dispatch({ type: LiveChatWidgetActionType.SET_MINIMIZED, payload: false });
+        } else {
+            await prepareStartChat();
+        }
+    };
```
elopezanaya, line 38: I feel like we could have this functionality definition in its own file, and just import it, reducing the complexity and size of this file

chat-widget/src/components/chatbuttonstateful/ChatButtonStateful.tsx
```diff
@@ -61,8 +65,8 @@ export const ChatButtonStateful = (props: IChatButtonStatefulParams) => {
                 dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.OutOfOffice });
             }
         },
-        unreadMessageString: props.buttonProps?.controlProps?.unreadMessageString,
-        ...outOfOfficeButtonProps?.controlProps
+        unreadMessageString: outOfOfficeChatButtonProps?.controlProps?.unreadMessageString,
+        ...outOfOfficeChatButtonProps?.controlProps
```
elopezanaya, line 69: I recommend we reconsidered this pattern for passing values and computing in-fly,  Im more in favor to have a function getWhatever Value and compute inside the function, in that way I have a single point of modification and do not mess with the component.

chat-widget/src/components/chatbuttonstateful/ChatButtonStateful.tsx
```diff
@@ -83,9 +87,9 @@ export const ChatButtonStateful = (props: IChatButtonStatefulParams) => {
 
     return (
         <ChatButton
-            componentOverrides={buttonProps?.componentOverrides}
+            componentOverrides={chatButtonProps?.componentOverrides}
             controlProps={outOfOperatingHours ? outOfOfficeControlProps : controlProps}
-            styleProps={outOfOperatingHours ? outOfOfficeStyleProps : buttonProps?.styleProps}
+            styleProps={outOfOperatingHours ? outOfOfficeStyleProps : chatButtonProps?.styleProps}
```
elopezanaya, line 92: what will happen if chatButtonProps is empty?, dont we need a default props?

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
```diff
@@ -132,4 +39,4 @@ const hasReconnectId = (reconnectAvailabilityResponse: IReconnectChatContext) =>
     return reconnectAvailabilityResponse && !isNullOrUndefined(reconnectAvailabilityResponse.reconnectId);
 };
 
-export { handleChatReconnect, isReconnectEnabled, isPersistentEnabled, getChatReconnectContext };
\ No newline at end of file
+export { redirectPage, isReconnectEnabled, isPersistentEnabled, hasReconnectId };
```
elopezanaya, line 42: my suggestion is to use : isReconnectIdPresent ?

chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
```diff
@@ -622,19 +620,19 @@ export const LiveChatWidgetStateful = (props: ILiveChatWidgetProps) => {
...
 
                         {!livechatProps.controlProps?.hideLoadingPane && shouldShowLoadingPane(state) && (decodeComponentString(livechatProps.componentOverrides?.loadingPane) || <LoadingPaneStateful loadingPaneProps={livechatProps.loadingPaneProps} startChatErrorPaneProps={livechatProps.startChatErrorPaneProps} />)}
 
                         {!livechatProps.controlProps?.hideOutOfOfficeHoursPane && shouldShowOutOfOfficeHoursPane(state) && (decodeComponentString(livechatProps.componentOverrides?.outOfOfficeHoursPane) || <OutOfOfficeHoursPaneStateful {...livechatProps.outOfOfficeHoursPaneProps} />)}
 
-                        {!livechatProps.controlProps?.hideReconnectChatPane && shouldShowReconnectChatPane(state) && (decodeComponentString(livechatProps.componentOverrides?.reconnectChatPane) || <ReconnectChatPaneStateful reconnectChatProps={livechatProps.reconnectChatPaneProps} initStartChat={initStartChatRelay} />)}
+                        {!livechatProps.controlProps?.hideReconnectChatPane && shouldShowReconnectChatPane(state) && (decodeComponentString(livechatProps.componentOverrides?.reconnectChatPane) || <ReconnectChatPaneStateful {...livechatProps} />)}
```
elopezanaya, line 633: I wonder if a more clean pattern will be to call a function that computes the state and return the Component when conditions are valid, instead of this way wich is more verbose

chat-widget/src/hooks/useCheckConversationValidity.ts
```diff
@@ -0,0 +1,58 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let conversationDetails: any = undefined;
+
+        //Preserve current requestId
+        const currentRequestId = chatSDK.requestId ?? "";
+        dispatch({ type: LiveChatWidgetActionType.SET_INITIAL_CHAT_SDK_REQUEST_ID, payload: currentRequestId });
+
+        try {
+            chatSDK.requestId = requestIdFromCache;
+            conversationDetails = await getConversationDetailsCall(chatSDK);
```
elopezanaya, line 29: if requestIdFromCache is null or empty, do we still need to call GetConversationDetail ?

chat-widget/src/hooks/useForceEndChat.ts
```diff
@@ -0,0 +1,21 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const chatSDK: any = useChatSDKStore();
+    const forceEndChat = () => {
+        TelemetryHelper.logLoadingEvent(LogLevel.ERROR, {
+            Event: TelemetryEvent.WidgetLoadFailed,
+            ExceptionDetails: {
+                Exception: "SessionInit was successful, but widget load failed."
+            }
+        });
+        chatSDK?.endChat();
```
elopezanaya, line 16: it seems like there is a chance where endChat is not executed , in the case of chatSDK being null... should we log that?

chat-widget/src/hooks/useSetCustomContext.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const getInitContextParamForPopoutFromOuterScope = async (scope: any): Promise<any> =>  {
+        let payload;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let waitPromiseResolve: any;
+        const waitPromise = new Promise((res, rej) => {
+            waitPromiseResolve = res;
+            setTimeout(() => rej("Failed to find method in outer scope"), 5000);
+        }).catch((rej) => console.warn(rej));
```
elopezanaya, line 28: console.error instead of warn ?

chat-widget/src/hooks/useSetCustomContext.ts
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const getInitContextParamForPopoutFromOuterScope = async (scope: any): Promise<any> =>  {
+        let payload;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        let waitPromiseResolve: any;
+        const waitPromise = new Promise((res, rej) => {
+            waitPromiseResolve = res;
+            setTimeout(() => rej("Failed to find method in outer scope"), 5000);
```
elopezanaya, line 27: why not configurable the 5 seconds wait? , maybe some customers want a shorter wait or longer...

elopezanaya (changes requested):

chat-widget/src/components/chatbuttonstateful/ChatButtonStateful.tsx
charliewang95: React hooks can only be called inside React components. So we'd rather add it here instead into another dummy component. 

The useRef is used here to counter the bug in FluentUI where onClick callback is not refreshed

chat-widget/src/hooks/useSetCustomContext.ts
charliewang95: This should not be error. By current design, even if the customer doesn't want to inject custom context we need to run this logic (to make sure there's no context from popout opener)

elopezanaya (approved):

chat-widget/src/components/chatbuttonstateful/ChatButtonStateful.tsx
```diff
@@ -1,49 +1,53 @@
...
-            if (state.appStates.isMinimized) {
-                dispatch({ type: LiveChatWidgetActionType.SET_MINIMIZED, payload: false });
-            } else {
-                await startChat();
-            }
-        },
-        unreadMessageString: props.buttonProps?.controlProps?.unreadMessageString,
-        ...buttonProps?.controlProps,
+        hideNotificationBubble: chatButtonProps?.controlProps?.hideNotificationBubble === true || state.appStates.isMinimized === false,
+        unreadMessageCount: state.appStates.unreadMessageCount ? (state.appStates.unreadMessageCount > Constants.maximumUnreadMessageCount ? chatButtonProps?.controlProps?.largeUnreadMessageString : state.appStates.unreadMessageCount.toString()) : "0",
```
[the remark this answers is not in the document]
charliewang95: Added to backlog

chat-widget/src/components/chatbuttonstateful/ChatButtonStateful.tsx
charliewang95: default props are defined in the component layer. Passing null here is acceptable

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
charliewang95: Renamed

chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
charliewang95: Added to backlog

chat-widget/src/hooks/useSetCustomContext.ts
charliewang95: This (custom context for popout) is a feature we need to revisit as a whole. Is not related to this PR

PR #337, opened 2023-08-09
Fix SSO magic code to use Web browser's version of BroadcastChannel

xTEddie:
- Replace usage of BroadcastChannel node package with browser's supported BroadcastChannel library for SSO magic code feature

### Test Scenarios
- SSO Magic code should be suppressed
- Magic Code page should not be opened
- Magic Code should be sent behind the scenes
- Magic Code should not appear in the chat widget

https://github.com/microsoft/omnichannel-chat-widget/assets/8888565/4d22f393-567d-4518-9e3d-e637166e059f

### DIFF
--- a/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
+++ b/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
@@ -3,9 +3,7 @@
 import { IStackStyles, Stack } from "@fluentui/react";
 import { LogLevel, TelemetryEvent } from "../../common/telemetry/TelemetryConstants";
 import React, { Dispatch, useEffect } from "react";
-
 import { BotMagicCodeStore } from "./webchatcontroller/BotMagicCodeStore";
-import { BroadcastChannel } from "broadcast-channel";
 import { Components } from "botframework-webchat";
 import { Constants } from "../../common/Constants";
 import { ILiveChatWidgetAction } from "../../contexts/common/ILiveChatWidgetAction";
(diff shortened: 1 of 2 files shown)


chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -54,8 +52,8 @@ const createMagicCodeSuccessResponse = (signin: string) => {
 export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
     const { BasicWebChat } = Components;
     const [state, dispatch]: [ILiveChatWidgetContext, Dispatch<ILiveChatWidgetAction>] = useChatContextStore();
-    const magicCodeBroadcastChannel = new BroadcastChannel(Constants.magicCodeBroadcastChannel);
-    const magicCodeResponseBroadcastChannel = new BroadcastChannel(Constants.magicCodeResponseBroadcastChannel);
+    const magicCodeBroadcastChannel = new (window as any).BroadcastChannel(Constants.magicCodeBroadcastChannel); // eslint-disable-line @typescript-eslint/no-explicit-any
```
elopezanaya, line 55: - was this test on pop-out?
- can we include the description and test evidence as part of the PR description?

elopezanaya (changes requested):

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
akhiloct1510: +1

xTEddie: - Pop-out is not a supported feature in this package. It's only in the OOB chat widget
- Added description and test evidences

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -3,9 +3,7 @@
 import { IStackStyles, Stack } from "@fluentui/react";
 import { LogLevel, TelemetryEvent } from "../../common/telemetry/TelemetryConstants";
 import React, { Dispatch, useEffect } from "react";
-
 import { BotMagicCodeStore } from "./webchatcontroller/BotMagicCodeStore";
-import { BroadcastChannel } from "broadcast-channel";
```
charliewang95, line 8: If this is not used any more, shall we remove it from package.json and yarn.lock?

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
charliewang95, line 55: Does window's Broadcast channel have all parity with the one we're using today?

charliewang95, line 55: Why are we creating new Broadcast channel like this instead of using `BroadcastServiceInitialize`? With this change won't there be 2 versions of BroadcastChannel existing at the same time?

xTEddie: It should because the Node package follows the Browser's API implementation. The main difference is that the Node package is more cross-platform.

xTEddie: `BroadcastServiceInitialize` is using the Node version of BroadcastChannel. It cannot be used since `MagicCodeForwarder.html` is using Browser's version of BroadcastChannel. So, the feature in `chat-widget` must use the Browser's version of BroadcastChannel only for the SSO Magic Code feature. Chat widget is still using the Node package version by default

charliewang95 (approved):

elopezanaya (approved):

PR #341, opened 2023-08-17
Fixed PreChatPane not showing full text for toggle input text

charliewang95:
[Bug 3546392](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3546392): [V2][ICM] Prechat with long lenght texts is not presented properly

### Description
PreChatPane not showing full text for toggle input text
image

### Solution Proposed
Added required styles to all ac-textBlock to show full text

### Acceptance criteria
Toggle input showing full text

### Test cases and evidence
image

### Sanity Tests

### A11y

### DIFF
--- a/chat-components/src/components/prechatsurveypane/PreChatSurveyPane.tsx
+++ b/chat-components/src/components/prechatsurveypane/PreChatSurveyPane.tsx
@@ -1,11 +1,11 @@
 import * as AdaptiveCards from "adaptivecards";
 
+import { ElementType, EventNames } from "../../common/Constants";
 import { IStackStyles, Stack } from "@fluentui/react";
 import React, { useCallback } from "react";
 import { addNoreferrerNoopenerTag, broadcastError, getInputValuesFromAdaptiveCard } from "../../common/utils";
 
 import { BroadcastService } from "../../services/BroadcastService";
-import { ElementType, EventNames } from "../../common/Constants";
 import { ICustomEvent } from "../../interfaces/ICustomEvent";
 import { IPreChatSurveyPaneProps } from "./interfaces/IPreChatSurveyPaneProps";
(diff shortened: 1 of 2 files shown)


chat-components/src/components/prechatsurveypane/PreChatSurveyPane.tsx
```diff
@@ -75,6 +75,8 @@ function PreChatSurveyPane(props: IPreChatSurveyPaneProps) {
                 font-size: ${props.styleProps?.customTextStyleProps?.fontSize} !important;
                 height: ${props.styleProps?.customTextStyleProps?.height};
                 padding-top: ${props.styleProps?.customTextStyleProps?.paddingTop};
+                overflow-wrap: break-word;
+                white-space: normal !important;
```
akhiloct1510, line 79: Can we avoid using Important ? Like may be defining css to more deeper level ?

charliewang95: AdaptiveCard has its own css defined at the same level. I don't see another way if we have to keep v1 parity

elopezanaya (approved):

akhiloct1510 (approved):

chat-components/src/components/prechatsurveypane/PreChatSurveyPane.tsx
charliewang95: AC has it's own css at the same level, which is defined later down the logic chain (which will always overwrite our styles). Technically we can add more parent selectors to remove the "!important", but that adds more dependencies (say a parent class name changes, we need to remember to change it here too, also more importantly the parent class names are all customizable). 

I think given that we need to keep V1 parity on this, using !important can be an acceptable exception

PR #345, opened 2023-08-28
Fix unauth reconnect bugs

charliewang95:
[Bug 3554753](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3554753): [V2] [8.3 PPE] [LCW] [UNAUTH] After the chat gets reconnected to previous chat, the error banner like "Start Chat Failed: TypeError: Cannot read properties of undefined (reading 'customContext')" is showing in the top of blob window.
[Bug 3552329](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3552329): [V2] Refreshing an active reconnect chat places chat in forever loading state

### Description
1. Refreshing unauth reconnect page results in infinite loading
2. Using reconnect id in url shows an error banner

## Solution Proposed
1. Detect when unauth reconnect page is being refreshed without reconnect id, then go with normal chat flow
2. reconnectHandler is missing an argument when doing initStartChat. Fixed the bug

### Acceptance criteria
Passed the following test cases and Reconnect regression suite
[Test Case 3561446](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3561446): [LCW OOB] [Reconnect Chat] Verify unauthenticated customer reconnects to chat after page refresh when third party cookie is on
[Test Case 2710644](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/2710644): [LCW OOB] [Reconnect Chat] Verify unauthenticated customer reconnects to chat using reconnect url within redirection time limit

## Test cases and evidence
proof1

### Sanity Tests
-   [x] You have tested all changes in Popout mode

## A11y

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
+++ b/chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
@@ -19,7 +19,7 @@ import StartChatOptionalParams from "@microsoft/omnichannel-chat-sdk/lib/core/St
 import { TelemetryHelper } from "../../../common/telemetry/TelemetryHelper";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-const handleChatReconnect = async (chatSDK: any, props: any, dispatch: Dispatch<ILiveChatWidgetAction>, setAdapter: any, initStartChat: any, state: ILiveChatWidgetContext) => {
+const handleChatReconnect = async (chatSDK: any, props: any, dispatch: Dispatch<ILiveChatWidgetAction>, setAdapter: any, initStartChat: any, state: ILiveChatWidgetContext): Promise<boolean | undefined> => {
     if (!isReconnectEnabled(props.chatConfig) || isPersistentEnabled(props.chatConfig)) return;
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -37,7 +37,7 @@ const handleChatReconnect = async (chatSDK: any, props: any, dispatch: Dispatch<
     if (hasReconnectId(reconnectChatContext)) {
         //if reconnect id is provided in props, don't show reconnect pane
         if (props.reconnectChatPaneProps?.reconnectId && !isNullOrEmptyString(props.reconnectChatPaneProps?.reconnectId)) {
-            await setReconnectIdAndStartChat(isAuthenticatedChat, chatSDK, props, dispatch, setAdapter, reconnectChatContext.reconnectId ?? "", initStartChat);
+            await setReconnectIdAndStartChat(isAuthenticatedChat, chatSDK, state, props, dispatch, setAdapter, reconnectChatContext.reconnectId ?? "", initStartChat);
             return;
         }
 
@@ -47,6 +47,10 @@ const handleChatReconnect = async (chatSDK: any, props: any, dispatch: Dispatch<
         dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.ReconnectChat });
         return;
     }
+
+    // If we have reached this point, it means there is no valid reconnect id or redirectUrl
+    // This is a unauth reconnect refresh scenario - returns true so that we can start normal hydration process
+    return true;
 };
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
```diff
@@ -160,9 +160,9 @@ export const LiveChatWidgetStateful = (props: ILiveChatWidgetProps) => {
             if (localState) {
                 // adding the reconnect logic for the case when customer tries to reconnect from a new browser or InPrivate browser
                 if (isReconnectEnabled(props.chatConfig) === true && !isPersistentEnabled(props.chatConfig)) {
-                    await handleChatReconnect(chatSDK, props, dispatch, setAdapter, initStartChat, state);
+                    const noValidReconnectId = await handleChatReconnect(chatSDK, props, dispatch, setAdapter, initStartChat, state);
                     // If chat reconnect has kicked in chat state will become Active or Reconnect. So just exit, else go next
-                    if (state.appStates.conversationState === ConversationState.Active || state.appStates.conversationState === ConversationState.ReconnectChat) {
+                    if (!noValidReconnectId && (state.appStates.conversationState === ConversationState.Active || state.appStates.conversationState === ConversationState.ReconnectChat)) {
                         return;
                     }
```
elopezanaya, line 167: this block seems to be the same as LoC 148 to LoC 151,  could we reuse it instead?

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
```diff
@@ -47,6 +47,10 @@ const handleChatReconnect = async (chatSDK: any, props: any, dispatch: Dispatch<
         dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.ReconnectChat });
         return;
     }
+
+    // If we have reached this point, it means there is no valid reconnect id or redirectUrl
+    // This is a unauth reconnect refresh scenario - returns true so that we can start normal hydration process
+    return true;
```
elopezanaya, line 53: I have conflicts with returning true for this point, while we return undefined for others, 

is not clear exactly what is the meaning of true, since the purpose of the function is to handle the reconnect, not to validate, and this "true" doesn't really mean the reconnect happened successfully.

could we considerate using ENUMS to pass the scenario reached?

by example, HandleReconnect.NovalidRecconnectIdScenario

elopezanaya (changes requested):

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
charliewang95: Using Enum for a helper function with boolean return value is over-engineering IMO. I've changed return type to boolean only and added comment for the return value's meaning

chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
charliewang95: refactored this

elopezanaya (approved):

chat-widget/src/components/livechatwidget/common/reconnectChatHelper.ts
```diff
@@ -31,22 +32,26 @@ const handleChatReconnect = async (chatSDK: any, props: any, dispatch: Dispatch<
...
-        return;
+        return false;
     }
 
     if (hasReconnectId(reconnectChatContext)) {
         //if reconnect id is provided in props, don't show reconnect pane
         if (props.reconnectChatPaneProps?.reconnectId && !isNullOrEmptyString(props.reconnectChatPaneProps?.reconnectId)) {
-            await setReconnectIdAndStartChat(isAuthenticatedChat, chatSDK, props, dispatch, setAdapter, reconnectChatContext.reconnectId ?? "", initStartChat);
-            return;
+            await setReconnectIdAndStartChat(isAuthenticatedChat, chatSDK, state, props, dispatch, setAdapter, reconnectChatContext.reconnectId ?? "", initStartChat);
```
kevinlin98z, line 41: Why are we introducing `state` as a parameter? Is it because we are including the scenario to set custom context?

charliewang95: "state" is required in initStartChat. It is a bug for omitting it here

kevinlin98z (approved):

nareshme2016 (approved):

PR #362, opened 2023-10-02
[Draft]Updated audio notification to handle failures

dishakhattri:
### Description

## Solution Proposed

### Acceptance criteria

## Test cases and evidence

### Sanity Tests

## A11y


dishakhattri:
Modern browsers dont allow the auto-play until there is a user interaction on the window. Try granting the autoplay permissionspolicy to grant the autoplay support to the iframe containing the livechatwidget. This may be blocking the playback and thus the error above. Please refer to: [https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide#:~:text=Autoplay availability,and its document Opens in new window or tab](https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide#:~:text=Autoplay%20availability,and%20its%20document). Since I am unable to reproduce this scenario, the DOMexception will be a bandaid fix which is not a good solution. This is certainly needs to be handled on the portal side.

PR #390, opened 2023-11-06
Adding authentication check to fetch new token before endchat

sarojkpr:
[Issue 3651972](https://dev.azure.com/dynamicscrm/OneCRM/_workitems/edit/3651972): [DFC] - Chat Survey return rate discrepancies in Consumer

[Incident 436007411](https://icmcdn.akamaized.net/imp/v3/incidents/details/436007411/home) : Post chat survey didn't load for long ongoing chats

### Description
**Issue:** The chat widget fails to return post-chat surveys for long-running chats, which has been reported by several SxG organization partners. This has resulted in a significant drop of approximately 25% in receiving feedback rates.

**Investigation Findings**: Upon investigating the issue, we identified that the problem is due to 401 unauthorized error returned from the `prepareEndChat` function.

**Consistency:** The issue appears to be consistently reproducible in the context of long-running chats.

## Solution Proposed
To address this issue, the recommendation is to implement a solution that ensures the `chatSDK` has a valid authentication token before executing the remaining code in `prepareEndChat` function, in case of authenticated chats. The included fix will get a new token and allow `chatSDK.getConversationDetails()` to run successfully, without encountering unauthorized errors, and subsequently, the survey can be rendered.

### Acceptance criteria
1. Ensure unauthenticated chats is working as expected.
2. Ensure authenitcated chats are working as expected without causing any regression
3. Ensure post chat survey is rendered successfully even for long running chats.

## Test cases and evidence

### Sanity Tests

## A11y
Not required, as there is no change to UX.

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/endChat.ts
+++ b/chat-widget/src/components/livechatwidget/common/endChat.ts
@@ -19,6 +19,9 @@ import { TelemetryManager } from "../../../common/telemetry/TelemetryManager";
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const prepareEndChat = async (props: ILiveChatWidgetProps, chatSDK: any, state: ILiveChatWidgetContext, dispatch: Dispatch<ILiveChatWidgetAction>, setAdapter: any, setWebChatStyles: any, adapter: any) => {
     try {
+        //Get new auth token again if chat continued for longer time, otherwise chatsdk throws 401 error for auth chats
+        await handleAuthenticationIfEnabled(props, chatSDK);
+
         // Use Case: If call is ongoing, end the call by simulating end call button click
         endVoiceVideoCallIfOngoing(chatSDK, dispatch);
 


chat-widget/src/components/livechatwidget/common/endChat.ts
```diff
@@ -19,6 +19,9 @@ import { TelemetryManager } from "../../../common/telemetry/TelemetryManager";
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const prepareEndChat = async (props: ILiveChatWidgetProps, chatSDK: any, state: ILiveChatWidgetContext, dispatch: Dispatch<ILiveChatWidgetAction>, setAdapter: any, setWebChatStyles: any, adapter: any) => {
     try {
+        //Get new auth token again if chat continued for longer time, otherwise chatsdk throws 401 error for auth chats
+        await handleAuthenticationIfEnabled(props, chatSDK);
```
charliewang95, line 23: Update CHANGE_LOG please

charliewang95, line 23: If token is not expired, will this hurt performance?

sarojkpr: Yes and No, 
Yes - because it's making an additional call to get token. 
No - because chat is already ending at this stage.

Should we allow this PR to fix the breaking functionality and have another tracking item for perf improvement?

sarojkpr: I was reviewing the code again, if performance is a problem, then instead of introducing a new `handleAuthenticationIfEnabled`, maybe we should move the existing one from `endChat` to this method in `prepareEndChat`?

PR #438, opened 2024-02-27
3841351 : Prechat pane issues with state handling

elopezanaya:
### Description

When a chat with prechat is displayed, the prechat is minimized and there is a reload, during the reload as part of internal logic the  reload takes from cache state that needs to display the pane , so it triggers the launch of the pane . but then the conversation is set to close and the button is presented.

when the button is presented, by clicking the button it launch the widget and present the pre chat again, but when the button is hidden and the chat is launched via startChat method, then is not launching the chat.

but when inspecting the DOM we can see the pane present, blocking elements behind.

- Another related issue is for multiple tabs, when minimizing the chat in one tab , block the other tabs to launch the tab via start chat method, because there is no syncronization of the minimize state among tabs.

- it was detected as well that the inMemory State component, is not in sync with minimize state, but returning old state.

## Solution Proposed

- when reloading , and conversation is in prechat mode, set the state to close, this will prevent to launch the pre chat pane and wait for start logic to present the pane.

- include a listener for a new method so the OOB can notify when the minimize happened in another tab , in this way the state will be updated and next time start chat is launched, the pre chat will be displayed.

### Acceptance criteria

- prechat pane is presented each time a start chat is trigered
- reload doesnt prevent prechat to be presented
-  multitab honors states and present pane

## Test cases and evidence

### Sanity Tests

## A11y

### DIFF
--- a/chat-widget/src/components/livechatwidget/common/startChat.ts
+++ b/chat-widget/src/components/livechatwidget/common/startChat.ts
@@ -76,10 +76,12 @@ const setPreChatAndInitiateChat = async (chatSDK: any, dispatch: Dispatch<ILiveC
         } else {
             dispatch({ type: LiveChatWidgetActionType.SET_PRE_CHAT_SURVEY_RESPONSE, payload: preChatSurveyResponse });
             dispatch({ type: LiveChatWidgetActionType.SET_CONVERSATION_STATE, payload: ConversationState.Prechat });
-            // If minimized, maximize the chat
+            
+            // If minimized, maximize the chat, if the state is missing, consider it as minimized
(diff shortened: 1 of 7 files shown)


nareshme2016 (approved):

chat-widget/src/components/livechatwidget/common/startChat.ts
```diff
@@ -129,7 +142,7 @@ const initStartChat = async (chatSDK: any, dispatch: Dispatch<ILiveChatWidgetAct
                 }
             });
         }
-
+
```
xTEddie, line 145: nit: remove space

xTEddie (approved):

PR #469, opened 2024-04-26
Support dynamic style customization for System messages

mithunjohny:
ID: https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3988316

### Description
Problem Statement:
SxG, when they switch the theme to dark mode dynamically, they are not able to change the colors of the system messages, so it is unreadable and does not pass accessibility tests.

Ask:
SxG has a fast-follow ask to be able to edit the system messages in the WebChat container "An agent will be with you in just a minute." dynamically. 

They want to be able to change the color, and size of the font for all system messages.

P0- is be able to different color and size of fonts to apply these settings for all the messages.

P1 - font family, font style

## Solution Proposed
Added new useEffect react hook in liveChatWidgetStatefull class which will monitor updates to props object. When 
SDK.updateCustomizationProperties is called with new props, newly added useEffect will be called and it will then call the dispatch method to update the state variable SET_RENDERING_MIDDLEWARE_PROPS

### Acceptance criteria
User able to dynamically update the system message color,fonts, font family, font style etc

## Test cases and evidence
https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/3922332

image

image

### Sanity Tests
-   [x] You have tested all changes in Popout mode
-   [x] You have tested all changes in cross browsers i.e Edge, Chrome, Firefox, Safari and mobile devices(iOS and Android)
-   [x] Your changes are included in the [CHANGELOG](../CHANGE_LOG.md)

## A11y

### DIFF
--- a/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
+++ b/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-import { IStackStyles, Stack } from "@fluentui/react";
+import { IStackStyles, Stack, IRawStyle } from "@fluentui/react";
 import { LogLevel, TelemetryEvent } from "../../common/telemetry/TelemetryConstants";
 import React, { Dispatch, useEffect } from "react";
 
@@ -203,8 +203,12 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
         .webchat__bubble:not(.webchat__bubble--from-user) .webchat__bubble__content {
             border-radius: 0 !important; /* Override border-radius */
         }
+
+        .webchat__stacked-layout_container>div {
+            background: ${(props?.webChatContainerProps?.containerStyles as IRawStyle)?.background?? ""}
+        }
         `}</style>
-        <Stack styles={containerStyles}>
+        <Stack styles={containerStyles} className="webchat__stacked-layout_container">
             <BasicWebChat></BasicWebChat>
         </Stack>
         </>
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -6,12 +6,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+
 ### Changed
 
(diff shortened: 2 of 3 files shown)


matan-hershberg-nuance (approved):

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -203,8 +203,12 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
         .webchat__bubble:not(.webchat__bubble--from-user) .webchat__bubble__content {
             border-radius: 0 !important; /* Override border-radius */
         }
+
+        .webchat__stacked-layout_container>div {
+            background: ${(props?.webChatContainerProps?.containerStyles as IRawStyle)?.background?? ""}
```
elopezanaya, line 208: when the props is empty, this will override any style, do we need to define a default style ?, in that case when props is not passed, then it will use default

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -203,8 +203,12 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
...
+        }
         `}</style>
```
elopezanaya, line 210: I wonder if this should be a better place to apply the change, it will honor the update of style each time there is a change

 https://github.com/elopezanaya/omnichannel-chat-widget/blob/0fa36a3533abfee2bdd9af7b07ec749de294a8a5/chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/activityMiddleware.tsx#L82-L83

elopezanaya (changes requested):

elopezanaya (approved):

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
mithunjohny: Leaving it empty because it will be overriden by the bot framework default background color.

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
mithunjohny: This is not for activity , this class is added for the transcript container background

PR #485, opened 2024-06-05
Logical gate to avoid multiple calls  to getLiveWorkItemDetails

elopezanaya:
4153717 & 4153744 
### Description

in current model, LCW each time the user switch focus among tabs , will call getLiveWorkItemdetails to corroborate the state of the conversation.

the problem relays on , the react component uses useEfffect when reloading the component to register the listener for this action, causing each time there is switch and launch the component again to register additional listener, 

in simple scenarios when customer do a couple of switches, multiple calls will be made, adding to the retry mechanism in oc-sdk, leads to the endpoint to return 400 HTTP Error code for all the calls.

each call will have 3 retries, so for an scenario of 3 switches, for the 4th switch it will perform 12 calls for the same endpoint,

image
image

## Solution Proposed

- logic gate to avoid multiple register of listeners
- logic gate to avoid multiple calls colliding

__this should reduce our calls to the endpoint for a total of N^3, where N is the number of swichs among tabs__

### Acceptance criteria

- calls to getLiveWorkItemDetails when switching should honor the rules
   - do not register additional listeners
   - one call at the time 
   - honor 3 minutes wait time between calls
   - NOT calls in the parent window

## Test cases and evidence

### embedded , single call when switching
image
### popout
image
### Sanity Tests

## A11y

### DIFF
--- a/chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
+++ b/chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
@@ -4,6 +4,8 @@ import { Components, StyleOptions } from "botframework-webchat";
 import { ConfirmationState, Constants, ConversationEndEntity, E2VVOptions, LiveWorkItemState, PrepareEndChatDescriptionConstants, StorageType } from "../../../common/Constants";
 import { IStackStyles, Stack } from "@fluentui/react";
 import React, { Dispatch, useEffect, useRef, useState } from "react";
+import { TelemetryManager, TelemetryTimers } from "../../../common/telemetry/TelemetryManager";
+import { chatSDKStateCleanUp, endChat, endChatStateCleanUp, prepareEndChat } from "../common/endChat";
 import { checkIfConversationStillValid, initStartChat, prepareStartChat, setPreChatAndInitiateChat } from "../common/startChat";
 import {
     createTimer,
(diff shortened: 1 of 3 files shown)


matan-hershberg-nuance (approved):

mithunjohny (approved):

chat-widget/src/components/livechatwidget/livechatwidgetstateful/LiveChatWidgetStateful.tsx
```diff
@@ -292,16 +294,29 @@ export const LiveChatWidgetStateful = (props: ILiveChatWidgetProps) => {
                     dispatch({ type: LiveChatWidgetActionType.SET_MINIMIZED, payload: event?.payload?.isChatHidden });
                 }
                 const dateNow = Date.now();
-                if (dateNow - lastLWICheckTimeRef.current > Constants.LWICheckOnVisibilityTimeout) {
-                    const conversationDetails = await getConversationDetailsCall(chatSDK);
+
+                if (window?.location?.href?.includes("open-in-window=true") && !window?.location?.href?.includes("is-popout-mode=true")) {
```
mithunjohny, line 298: A suggestion, can we move this to a util class and use function reference like 
isPopuOutChatMode

elopezanaya: done

PR #486, opened 2024-06-05
Transcript External Scripts Support

xTEddie:
Task #3945805

### Description
- Add ability to pass custom CDN URLs of the dependencies from the transcript feature

## Test cases and evidence
- Validated manually the transcript feature works fine with the custom CDN URLs
- Validated manually the feature still works fine without custom CDN URLs

Transcript Custom CDNs

### DIFF
--- a/chat-widget/src/plugins/createChatTranscript.ts
+++ b/chat-widget/src/plugins/createChatTranscript.ts
@@ -1,39 +1,7 @@
 /* eslint-disable no-useless-escape */
 
 import { createFileAndDownload } from "../common/utils";
-
-interface TranscriptHTMLScripts {
-    botframeworkWebChat?: {
-        src?: string;
-        integrity?: string;
-        crossOrigin?: string;
-        referrerPolicy?: string;
-    },
-    rxJs?: {
-        src?: string;
-        integrity?: string;
-        crossOrigin?: string;
-        referrerPolicy?: string;
-    },
-    react?: {
-        src?: string;
-        integrity?: string;
-        crossOrigin?: string;
-        referrerPolicy?: string;
-    },
-    reactDom?: {
-        src?: string;
-        integrity?: string;
-        crossOrigin?: string;
-        referrerPolicy?: string;
-    },
-    markdownIt?: {
-        src?: string;
-        integrity?: string;
-        crossOrigin?: string;
-        referrerPolicy?: string;
-    }
-}
+import TranscriptHtmlScripts from "../components/footerstateful/downloadtranscriptstateful/interfaces/TranscriptHtmlScripts";
 
 const defaultLibraryScripts = {
     botframeworkWebChat: {
@@ -71,7 +39,7 @@ class TranscriptHTMLBuilder {
     private customerAvatarFontColor = "#FFF";
     private disableMarkdownMessageFormatting = false;
     private disableNewLineMarkdownSupport = false;
-    private externalScripts: TranscriptHTMLScripts = {};
+    private externalScripts: TranscriptHtmlScripts = {};
 
     constructor(options: any) {  // eslint-disable-line @typescript-eslint/no-explicit-any
         this.options = options;
--- a/chat-widget/src/plugins/createChatTranscript.ts
+++ b/chat-widget/src/plugins/createChatTranscript.ts
@@ -1,30 +1,8 @@
 /* eslint-disable no-useless-escape */
 
 import { createFileAndDownload } from "../common/utils";
+import defaultLibraryScripts from "../components/footerstateful/downloadtranscriptstateful/common/defaultLibraryScripts";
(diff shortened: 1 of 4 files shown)


chat-widget/src/plugins/createChatTranscript.ts
```diff
@@ -2,6 +2,62 @@
 
 import { createFileAndDownload } from "../common/utils";
 
+interface TranscriptHTMLScripts {
```
elopezanaya, line 5: my reccomendation is to extract this to an external file

chat-widget/src/plugins/createChatTranscript.ts
```diff
@@ -80,13 +141,33 @@ class TranscriptHTMLBuilder {
         return htmlData;
     }
 
+    createScriptElement(src: string, integrity: string | undefined = undefined, crossOrigin: string | undefined = undefined, referrerPolicy: string | undefined = undefined) {
+        return `<script src="${src}" ${integrity ? `integrity="${integrity}"`: ""} ${crossOrigin ? `crossorigin="${crossOrigin}"`: ""} ${referrerPolicy ? `referrerpolicy="${referrerPolicy}"`: ""}><\/script>`;
+    }
+
     createExternalScriptElements() {
+        const webChatScript = this.externalScripts?.botframeworkWebChat?.src?
+            this.createScriptElement(this.externalScripts?.botframeworkWebChat?.src as string, this.externalScripts?.botframeworkWebChat?.integrity, this.externalScripts?.botframeworkWebChat?.crossOrigin, this.externalScripts?.botframeworkWebChat?.referrerPolicy):
```
elopezanaya, line 150: same, lets keep in mind single responsability principle, lets abstract this functionality , aiming for readability and maintenablity

elopezanaya (changes requested):
request for cleanest/modular implementation

chat-widget/src/plugins/createChatTranscript.ts
```diff
@@ -80,13 +141,33 @@ class TranscriptHTMLBuilder {
         return htmlData;
     }
 
+    createScriptElement(src: string, integrity: string | undefined = undefined, crossOrigin: string | undefined = undefined, referrerPolicy: string | undefined = undefined) {
+        return `<script src="${src}" ${integrity ? `integrity="${integrity}"`: ""} ${crossOrigin ? `crossorigin="${crossOrigin}"`: ""} ${referrerPolicy ? `referrerpolicy="${referrerPolicy}"`: ""}><\/script>`;
+    }
+
     createExternalScriptElements() {
+        const webChatScript = this.externalScripts?.botframeworkWebChat?.src?
```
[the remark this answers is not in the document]
xTEddie: I agree this way would be cleaner, but it won't cover the scenario when `externalScript?.src` is `""`. `src` property **MUST** always be valid. Otherwise, transcript won't have all the necessary scripts to be functional.

I am still open for suggestions to on how this can be simplified for better readability

chat-widget/src/plugins/createChatTranscript.ts
```diff
@@ -1,6 +1,30 @@
...
+    },
+    reactDom: {
+        src: "https://unpkg.com/react-dom@18.2.0/umd/react-dom.production.min.js"
+    },
+    markdownIt: {
+        src: "https://cdn.jsdelivr.net/npm/markdown-it@13.0.1/dist/markdown-it.min.js",
+        integrity: "sha256-hNyljag6giCsjv/yKmxK8/VeHzvMDvc5u8AzmRvm1BI=",
+        crossOrigin: "anonymous",
+    }
+};
```
elopezanaya, line 27: the current pattern for any default definition is in a separate component,

chat-widget/src/plugins/createChatTranscript.ts
```diff
@@ -80,13 +109,53 @@ class TranscriptHTMLBuilder {
         return htmlData;
     }
 
+    createScriptElement(src: string, integrity: string | undefined = undefined, crossOrigin: string | undefined = undefined, referrerPolicy: string | undefined = undefined) {
+        return `<script src="${src}" ${integrity ? `integrity="${integrity}"`: ""} ${crossOrigin ? `crossorigin="${crossOrigin}"`: ""} ${referrerPolicy ? `referrerpolicy="${referrerPolicy}"`: ""}><\/script>`;
+    }
```
elopezanaya, line 114: I wonder if we should not include the script attribute that is not defined, or we should pick default when undefined

elopezanaya (changes requested):
one ask to move to its own component the default definition, open to discuss over the script generation output

chat-widget/src/plugins/createChatTranscript.ts
xTEddie: Validations are already done in `createExternalScriptElements()` to ensure `src` is defined. Otherwise, the default script is used.

elopezanaya (approved):

mithunjohny (approved):

PR #564, opened 2024-12-17
export createAttachmentMiddleware to reduce changes for custom middlewares provided by hosts

bkebede252:
### Description
- We were trying to create a custom attachmentmiddleware, it seems that if we copy the current code we'd miss out on future implementations. By exporting the createAttachmentmiddleware, we can wrap custom code on top of the default attachmentmiddleware. Here is a sample code :

import { createAttachmentMiddleware } from '@microsoft/omnichannel-chat-widget'
  
const customAttachmentMiddleware =
      () =>
      () =>
      (next: any) =>
      (...args: any) => {
        // set custom code here to return custom component or logic before triggering default attachment middleware
        const [card] = args
        if (!card?.activity) {
          return next(...args)
        }

      const {
        activity: { attachments },
        attachment,
      }: { activity: { attachments: any }; attachment: any } = card
      // No attachment
      if (!attachments || !attachments.length || !attachment) {
        return next(...args)
      }

      // Check for Adaptive cards
      // eslint-disable-next-line prefer-const
      let { content, contentType } = attachment || { content: '', contentType: '' }
      let { type } = content || { type: '' }

      if (type === 'AdaptiveCard' || contentType === 'application/vnd.microsoft.card.adaptive') { 
        return <SampleComponent attachment={ attachment }/>
      }

      // end of custom code 
      // default attachment middleware from omnichannel-chat-widget
      return createAttachmentMiddleware(enableInlinePlaying)()(next)(...args)
    }

## Solution Proposed
- Export createAttachmentMiddleware

### Acceptance criteria

## Test cases and evidence

### Sanity Tests

## A11y

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -20,7 +20,7 @@ All notable changes to this project will be documented in this file.
 - Padding property to control the padding size choice input adaptive card form field
 
 ### Changed
-
+- export createAttachmentMiddleware to be imported by hosts
 - Uptake [@microsoft/omnichannel-chat-sdk@1.10.2](https://www.npmjs.com/package/@microsoft/omnichannel-chat-sdk/v/1.10.2)
 - Uptake [@opentelemetry/api@1.9.0](https://www.npmjs.com/package/@opentelemetry/api/v/1.9.0)
 - Uptake [@azure/core-tracing@1.2.0](https://www.npmjs.com/package/@azure/core-tracing/v/1.2.0)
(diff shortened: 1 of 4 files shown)


bkebede252:
@microsoft-github-policy-service agree company="Microsoft"

mithunjohny (approved):

elopezanaya (changes requested):
This is a particular solution for a particular case, that doesnt fit a global approach.

- instead of exposing just one middleware we should discuss about exposing all of them and how to control telemetry
- another alternative is to provide a registry for external middleware , were externals can hook their own callbacks, 

Exposing just 1 middleware I dont think is a good strategy, lets discuss and provide a design for a global approach..

elopezanaya:
to be discussed by the team

PR #727, opened 2025-07-22
Broadcast Service Event Queue Improvements

xTEddie:
### Description
- Event Queue was getting processed at an interval which is not performant enough

## Solution Proposed
- Improvement of the Event Queue
  - Event Queue now does not process the events at an interval
  - Queue will stop processing as soon as queue size is empty
  - `deferTimeout` was removed given the new logic would not cause any race condition

## Acceptance criteria
Event queue is working as expected and not getting processed at interval
 
## Test cases and evidence
- Manually validation with the sample app was done

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -599,6 +599,10 @@ All notable changes to this project will be documented in this file.
 
 - Add `EventQueue` to emit un-processed events as fallback to `BroadcastService`
 
+### Changed
+
+- Update `EventQueue` to not process events at interval by stopping queueing as soon as queue size is empty
+
 ## [1.1.12] - 2025-05-29
 
 ### Changed


chat-components/src/services/EventQueue.ts
```diff
@@ -12,20 +12,20 @@ class EventQueue {
         this.newMessage = newMessage;
     }
 
-    processEvents(deferTimeout = 0) {
+    processEvents() {
         const dequeue = () => {
```
elopezanaya, line 16: why we need dequeue function definition if you are calling it right after

chat-components/src/services/EventQueue.ts
```diff
@@ -12,20 +12,20 @@ class EventQueue {
...
-        setTimeout(() => {
-            dequeue();
-        }, deferTimeout);
+        dequeue();
     }
 
     queueEvents(timeout = 500) {
+        this.stopIfEmpty();
+
         if (this.channelEventQueue.size === 0) { // Base case
```
mithunjohny, line 29: no longer needed right

chat-components/src/services/EventQueue.ts
```diff
@@ -12,20 +12,20 @@ class EventQueue {
...
         };
 
-        setTimeout(() => {
-            dequeue();
-        }, deferTimeout);
+        dequeue();
     }
 
     queueEvents(timeout = 500) {
+        this.stopIfEmpty();
```
elopezanaya, line 27: isnt this duplicate, stopIfEmpty is already doing this validation

xTEddie: Logic removed

mithunjohny (approved):

elopezanaya (approved):

PR #923, opened 2026-05-08
fix(a11y): make citation a11y patch idempotent to break MutationObserver loop

dandriscoll:
> ⚠️ **Stacked on #918** — the diff currently includes the accessibility-tooling foundation commits. Once #918 merges this PR will rebase down to just the citation-card fix.

## Bug
The `MutationObserver` in `WebChatContainerStateful` re-invokes `patchCitationAnchorsForA11y` on every `aria-hidden` / `role` / `tabindex` / `inert` mutation. The patch unconditionally `setAttribute`'d those same attributes on every descendant, which triggered the observer again, looping indefinitely and freezing the page under load (especially with long bot transcripts that include citations).

## Fix
Make the patch idempotent: only `setAttribute` when the desired value differs from the current value, so the MutationObserver no longer fires on a no-op write. The observer and patch are scoped to the WebChat root and citation-card anchors so host-page and inline `cite:` links are not mutated.

## Tests
Adds unit coverage for idempotent re-patching and for preserving bare `cite:` links outside citation cards.

---


chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
```diff
@@ -0,0 +1,225 @@
...
+//  - Add a transparent ::after pseudo-element that covers the entire anchor.
+//    iOS Safari hit-tests that overlay as a SINGLE region, so the tap
+//    highlight / selection rectangle always covers the whole card instead
+//    of just the "1" badge or just the link text.
+//  - Make every descendant fully transparent to pointer / selection / focus,
+//    so they can never become a standalone tap target.
+const STYLE_RULES = `
+a.webchat__link-definitions__list-item-box,
+a[href^="cite:"] {
+    position: relative !important;
```
MugdhaSMundhe: Could we consider dropping the !important flags from the 16 declarations in these two blocks? Once the selectors are scoped to the chat-widget root, specificity alone should be sufficient — and removing !important would let host-page themes and consumer overrides continue to work as expected.

chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
```diff
@@ -0,0 +1,225 @@
...
+    display: block !important;
+    isolation: isolate !important;
+    -webkit-tap-highlight-color: rgba(0, 120, 212, 0.15) !important;
+    -webkit-touch-callout: none !important;
+    -webkit-user-select: none !important;
```
MugdhaSMundhe: In the same blocks, would you also be open to removing user-select: none / -webkit-user-select: none? Right now sighted mouse and keyboard users lose the ability to select and copy citation title text, which I think is an unintended side effect of the iOS VoiceOver fix. Keeping selection enabled on the title would preserve that behavior for everyone else.

MugdhaSMundhe (approved):

chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
dandriscoll: Agreed. I’ll remove the unnecessary !important declarations so the scoped widget styles don’t block host/theme overrides.

chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
dandriscoll: Agreed. I’ll keep citation title text selectable and remove the user-select suppression from this fix.

chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
dandriscoll: Addressed in prior commit.

chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
dandriscoll: Addressed in prior commit.

BJohnsonMicrosoft (approved):

jatinjd2904 (commented):
Mostly solid  the idempotency fix itself is correct and well-tested. A few things worth thinking about before this lands, mostly around scope and side-effects.

**1. The PR is much bigger than the title.**
Title is "make citation a11y patch idempotent to break MutationObserver loop," but the diff is +537 lines and brings in the entire ``citationA11y.ts`` utility, the ``WebChatContainerStateful`` MutationObserver wiring, the global CSS injector, and the DOM overlay span. I get that it's stacked on #918  please call out very visibly in the description what the *actual* delta is once #918 merges, otherwise reviewers (and approvers above) end up signing off on a much larger surface than the title implies. Mugdha's approval is on a commit that didn't even include the "preserve style overrides" change.

**2. ``display`` clobber on host themes.**
In ``citationA11y.ts``:
```ts
if (!anchorStyle.display || anchorStyle.display === "inline") {
    anchorStyle.display = "block";
}
```
``anchorStyle.display`` reads the **inline** style, not the computed style. If a host stylesheet/class sets ``display: flex`` (or grid) on the citation anchor, ``anchorStyle.display`` is ``""``, which is falsy, so we set ``display: block`` inline  and inline beats class. Same risk for ``position``, though ``position: relative`` is usually a safe default. Suggest using ``getComputedStyle`` before clobbering, or only setting if the computed value is ``inline``/``static``.

**3. Global ``MutationObserver`` on ``document.body`` with ``subtree: true`` is heavy.**
The observer runs for the entire lifetime of the chat widget and fires on every ``aria-hidden``/``role``/``tabindex``/``inert`` change *anywhere on the host page*  not just inside WebChat. Each event triggers a ``closest(CITATION_ANCHOR_SELECTOR)`` walk. On busy host pages (animations, third-party widgets, dynamic dashboards) this is meaningful background work. Scoping the observer to the WebChat container element (or the live chat root) would cut this down enormously. If there's a reason it has to be ``document.body``, please drop a comment.

**4. The ``cite:`` selector is global and matches host-page anchors too.**
Both the injected CSS (``a[href^="cite:"]``) and the JS scan match *any* anchor on the page whose ``href`` starts with ``cite:``. If a host page legitimately uses ``cite:`` links (rare but valid for academic / footnote use), they'll get the inline ``pointer-events: none`` lockdown and the overlay. Worth scoping the selector to a known WebChat ancestor (``.webchat__... a[href^="cite:"]``) at least in the CSS, and walking from the WebChat root rather than ``document``.

**5. Double application of ``pointer-events: none``.**
The injected stylesheet sets ``a[href^="cite:"] * { pointer-events: none }``, and the JS *also* writes ``inner.style.pointerEvents = "none"`` on every descendant. Pick one  the inline write defeats CSS overridability for anyone debugging or theming, and the CSS rule already wins specificity. Inline-style writes are also extra observable mutations the observer has to chew through.

**6. ``title``  ``data-ocw-original-title`` is a behavior change worth a release note.**
Clever workaround for iOS VO's title-swipe behavior, but anything (third-party tooltip libraries, host CSS like ``[title]:hover``, screenshot tools) that reads ``title`` on inner descendants will silently stop working. Worth a one-liner in release notes or in the changelog entry beyond the current "single stable accessible link label" wording.

**7. Test isolation: injected ``<style>`` tag isn't cleaned up between tests.**
``afterEach`` only clears ``document.body.innerHTML``; the ``#ocw-citation-a11y-styles`` tag in ``document.head`` survives. Today the dedup guard hides this, but a future test that asserts on style content will see leakage from a previous test. One-line fix: ``document.getElementById('ocw-citation-a11y-styles')?.remove()`` in ``afterEach``.

**8. No regression test for the actual MutationObserver loop.**
``re-applies aria-hidden when stripped`` covers idempotency on the helper, but there's no test that wires up the helper + a real ``MutationObserver`` and asserts the loop terminates. The PR title is about the loop  would be nice to have a fixture-style test that runs the helper inside a JSDOM observer and asserts a finite number of callbacks. The "un-skipped catcher in #918" is fine but lives somewhere reviewers can't see in this PR.

**9. Tiny:** ``role="link"`` is only set when ``role`` is missing. Fine in practice, but if WebChat ever ships ``role="presentation"`` on the anchor itself (unlikely, but possible), we'd silently keep that and lose the single-link semantic. Either log or override is reasonable.

Net: (1) is purely description hygiene, (2) and (3) are the ones I'd want addressed before merge  both are real performance/regression hazards in a hot path. The rest is cleanup. Nice work overall  the test suite is genuinely good.

jatinjd2904 (approved):

chat-widget/src/components/webchatcontainerstateful/common/utils/citationA11y.ts
```diff
@@ -0,0 +1,221 @@
...
+//    highlight / selection rectangle always covers the whole card instead
+//    of just the "1" badge or just the link text.
+//  - Make every descendant transparent to pointer / focus, so they can never
+//    become a standalone tap target.
+const STYLE_RULES = `
+a.webchat__link-definitions__list-item-box,
+a[href^="cite:"] {
+    position: relative;
+    display: block;
+    isolation: isolate;
```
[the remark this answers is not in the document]
dandriscoll: Fixed by limiting the patch to citation-card anchors only; bare cite links are now covered by a preservation test.

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -159,6 +160,56 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
...
+    // Accessibility: Merge citation card's number badge and link text into a single
+    // focusable element. WebChat's LinkDefinitionItem renders an <a> containing a
+    // badge <div> (the number, e.g. "1") and a text <div> (the title). On iOS
+    // VoiceOver and Android TalkBack, those block-level descendants are otherwise
+    // announced as two separate focusable links. The patch sets a combined
+    // aria-label on the anchor and hides every descendant from the a11y tree so
+    // the whole card is announced as one link.
+    useEffect(() => {
+        patchCitationAnchorsForA11y(document);
+
```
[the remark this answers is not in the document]
dandriscoll: Fixed by scoping initial patching and the observer to #ms_lcw_webchat_root. Broader DOM patcher scoping is tracked in #936.

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -159,6 +160,56 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
...
+                }
+            }
+        });
+
+        observer.observe(document.body, {
+            childList: true,
+            subtree: true,
+            attributes: true,
+            attributeFilter: ["aria-hidden", "role", "tabindex", "inert"]
+        });
```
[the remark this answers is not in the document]
dandriscoll: Fixed by observing the web chat root instead of document.body. Broader DOM patcher scoping is tracked in #936.

CHANGE_LOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Fixed
+- [A11y] Citation cards now expose a single stable accessible link label and avoid duplicate title announcements
```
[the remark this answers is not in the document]
dandriscoll: Fixed by updating the PR description so it no longer claims the skipped Playwright catcher is re-enabled.

jatinjd2904 (approved):

VijVerma (approved):

PR #924, opened 2026-05-08
fix(a11y): merge adjacent markdown anchors with same href into single focusable link

dandriscoll:
> ⚠️ **Stacked on #918** — the diff currently includes the accessibility-tooling foundation commits. Once #918 merges this PR will rebase down to just the markdown-anchor merge fix.

## Bug

Bot responses such as `[1.](url) [View details](url)` rendered as two sibling `<a>` tags pointing at the same href. Screen readers announced two separate focusable links, confusing users of assistive technology.

## Fix

Post-process the markdown-it output to combine consecutive `<a>` elements that share an identical href into a single anchor, preserving inner content and whitespace so the accessible name reads naturally (e.g. `1. View details`).

Anchors with different hrefs are left untouched.

## Tests

Updates `createMarkdown.test.ts` with cases covering: matching href merge, mismatched href passthrough, and whitespace preservation.

---


chat-widget/src/components/livechatwidget/common/createMarkdown.ts
```diff
@@ -1,134 +1,158 @@
...
+                /(<a\b[^>]*?\shref="([^"]*)"[^>]*>)([\s\S]*?)<\/a>(\s*)<a\b[^>]*?\shref="\2"[^>]*>([\s\S]*?)<\/a>/gi;
+
+            let prev = "";
+            let curr = html;
+            // Loop to collapse runs of 3+ adjacent anchors (each pass merges pairs).
+            while (prev !== curr) {
+                prev = curr;
+                curr = curr.replace(anchorPairRegex, (_m, open1: string, _href: string, inner1: string, between: string, inner2: string) => {
+                    const separator = between && between.length > 0 ? between : " ";
+                    return `${open1}${inner1}${separator}${inner2}</a>`;
```
MugdhaSMundhe: Only open1 (the first anchor's opening tag) is preserved. Any attributes that differ on the second <a> — title, aria-label, rel, target, class, data-* — are discarded. The regex matches on href only, so two anchors with the same href but different aria-label/title (a real possibility when a bot composes "1. (url)" + "View details (url)") will lose the accessible-label override from the second tag. Consider either (a) refusing to merge when non-href attributes differ, or (b) explicitly merging 
 aria-label/title text rather than dropping them.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
```diff
@@ -1,134 +1,158 @@
...
+        // Accessibility fix: when a bot emits content like "[1.](url) [View details](url)",
+        // markdown-it renders two sibling <a> tags that screen readers announce as two
+        // separate focusable links. Merge consecutive anchors with identical href into a
+        // single <a> so the number and label form one combined, focusable link.
+        function mergeAdjacentAnchors(html: string): string {
+            // Matches: <a ... href="X" ...>inner1</a>[optional whitespace]<a ... href="X" ...>inner2</a>
+            // - backref \2 enforces identical href
+            // - [\s\S]*? keeps inner content minimal; markdown-it does not nest anchors
+            const anchorPairRegex =
+                /(<a\b[^>]*?\shref="([^"]*)"[^>]*>)([\s\S]*?)<\/a>(\s*)<a\b[^>]*?\shref="\2"[^>]*>([\s\S]*?)<\/a>/gi;
```
MugdhaSMundhe: Parsing HTML with regex is fragile, and this runs unconditionally on every md.render / md.renderInline call. Two concerns: (a) the href="..." pattern only handles double-quoted attributes — single-quoted or unquoted hrefs from html_inline passthrough won't match and won't be merged; (b) the back-reference \2 does a literal string compare, so semantically-identical but textually-different hrefs (e.g. trailing slash, URL-encoded vs decoded) won't merge. A fast-path if (!html.includes("</a>")) would also avoid the regex cost on the common no-anchor case.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
```diff
@@ -1,134 +1,158 @@
...
+                });
+            }
+            
+            return result;
+        }
+        
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        md.render = function(text: string, env?: any): string {
+            const processedText = preprocessText(text);
+            return mergeAdjacentAnchors(originalRender(processedText, env));
```
MugdhaSMundhe: The url_new_win plugin below (line 678) splices the open-in-new-window icon  into each anchor's inner content via tokens.splice(idx + 2, 0, ...iconTokens). After  mergeAdjacentAnchors runs, two adjacent same-href anchors both carry their own icon, and the merge concatenates inner1 + separator + inner2 — so the resulting single anchor renders two open-link icons in a row (e.g., 1. View details ). Please add a test for the disableMarkdownMessageFormatting=false path with two same-href adjacent links and de-duplicate the icon during merge.

MugdhaSMundhe (approved):

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
dandriscoll: Agreed. Merging only by href is too lossy if the second anchor carries distinct attributes or accessible-name metadata. I’ll adjust this so we don’t discard meaningful attributes.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
dandriscoll: Agreed. I’ll avoid relying on this broad regex merge path and make the behavior safer for non-identical anchor markup.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
dandriscoll: Agreed. I’ll add coverage for the new-window icon path and ensure the merge does not produce duplicate open-link icons.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
dandriscoll: Addressed in prior commit.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
dandriscoll: Addressed in prior commit.

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
dandriscoll: Addressed in prior commit.

jatinjd2904 (approved):

BJohnsonMicrosoft (approved):

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
```diff
@@ -1,134 +1,252 @@
...
+                        ...children.slice(index + 1, firstLink.closeIndex)
+                    ];
+                    let nextIndex = firstLink.closeIndex + 1;
+                    let mergedAnyLink = false;
+
+                    let nextLink = getNextAdjacentLink(children, nextIndex);
+                    while (nextLink
+                        && nextLink.href === firstLink.href
+                        && hasSameAttributes(firstLink.open, nextLink.open)) {
+                        linkTokens.push(...(nextLink.separatorTokens.length > 0 ? nextLink.separatorTokens : [createSpaceToken()]));
```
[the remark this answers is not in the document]
dandriscoll: Fixed: same-href link merging now preserves the original separator exactly and does not synthesize whitespace. Broader markdown transform invariants are tracked in #940.

jatinjd2904 (approved):

MSSunnyNayak (approved):

PR #926, opened 2026-05-08
fix(a11y): drop redundant aria-labelledby on compact ChoiceSet selects

dandriscoll:
> ⚠️ **Stacked on #918** — the diff currently includes the accessibility-tooling foundation commits. Once #918 merges this PR will rebase down to just the dropdown double-label fix.

## Bug

Compact `Input.ChoiceSet` selects rendered with **both** an `aria-labelledby` pointing at the field label and a wrapping `<label for>` referencing the same select. The duplicate naming caused screen readers (NVDA, JAWS) to announce the field label twice.

## Fix

Drop the redundant `aria-labelledby` so each compact ChoiceSet is named exactly once.

## Tests

Un-skips the `dropdown-double-label` catcher in `AdaptiveCardAccessibilityWrapper.unfixed.a11y.spec.tsx` (added by #918).

---


ashishkadam0 (approved):

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
```diff
@@ -28,6 +28,24 @@ const AdaptiveCardAccessibilityWrapper: React.FC<{
...
+            // <label for="..."> targeting the same select. Screen readers
+            // announce the same label twice. When a visible <label for>
+            // already associates with the select, drop the redundant
+            // aria-labelledby so only the native label association is
+            // announced.
+            const compactSelects = container.querySelectorAll<HTMLSelectElement>("select[aria-labelledby]");
+            compactSelects.forEach((select) => {
+                const id = select.id;
+                if (!id) return;
+                const visibleLabel = container.querySelector<HTMLLabelElement>(`label[for="${CSS.escape(id)}"]`);
```
MugdhaSMundhe: The presence of a <label for> is treated as sufficient justification to drop aria-labelledby, but the two are not compared. If aria-labelledby references a composite name (multiple IDs — e.g., field label + "required" indicator + inline error) or points at a richer element than the visible <label for>, dropping it silently loses the supplemental name. Please verify the IDs in aria-labelledby actually resolve to the same element as the visible label (or to a strict subset of its text) before removing the attribute — otherwise gate the strip on aria-labelledby containing exactly one token equal to the visible label's id.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
```diff
@@ -28,6 +28,24 @@ const AdaptiveCardAccessibilityWrapper: React.FC<{
...
 
+            // dropdown-double-label: compact Input.ChoiceSet renders as a native
+            // <select> AND emits BOTH `aria-labelledby` (pointing at the
+            // adaptive-card label element) AND a sibling visible
+            // <label for="..."> targeting the same select. Screen readers
+            // announce the same label twice. When a visible <label for>
+            // already associates with the select, drop the redundant
+            // aria-labelledby so only the native label association is
+            // announced.
+            const compactSelects = container.querySelectorAll<HTMLSelectElement>("select[aria-labelledby]");
```
MugdhaSMundhe: This matches every <select aria-labelledby> inside the wrapper, not just compact Input.ChoiceSet ones. Any other native <select> rendered in the card (custom HTML actions, future AC elements, third-party renderers) with an intentional aria-labelledby will lose its association. Scope the query to the compact ChoiceSet markup (e.g., select.ac-input.ac-multichoiceInput, or check that the parent matches .ac-input-container with a sibling <label for>) so the fix only fires on the known offender.

MugdhaSMundhe (approved):

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
dandriscoll: Agreed. I’ll only remove aria-labelledby when it truly duplicates the visible label and won’t drop composite required/error/context text.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
dandriscoll: Agreed. I’ll scope this to compact ChoiceSet selects instead of every select under the wrapper.

dandriscoll: Addressed in prior commit.

jatinjd2904 (commented):
_(superseded by the inline comment below)_

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
```diff
@@ -28,6 +28,23 @@ const AdaptiveCardAccessibilityWrapper: React.FC<{
...
+            // <select> with both aria-labelledby and a sibling visible <label for>.
+            // Only remove aria-labelledby when it points solely at that same visible
+            // label; preserve composite labels that include required/error/context text.
+            const compactSelects = container.querySelectorAll<HTMLSelectElement>(
+                ".ac-input-container select.ac-input.ac-multichoiceInput[aria-labelledby]"
+            );
+            compactSelects.forEach((select) => {
+                const id = select.id;
+                if (!id) return;
+                const visibleLabel = container.querySelector<HTMLLabelElement>(`label[for="${CSS.escape(id)}"]`);
```
jatinjd2904: Nit: this lookup doesn’t filter `aria-hidden`, but the unfixed-spec detector for this same scenario explicitly uses `:not([aria-hidden='true'])` when looking for the visible label. If Adaptive Cards ever emits a hidden `<label for>`, screen readers ignore it and `aria-labelledby` becomes the only accessible name — we’d then strip it and leave the select unnamed. Either match the spec’s filter here, or add a one-liner noting we’re assuming any `<label for>` is AT-visible.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
dandriscoll: Fixed: aria-labelledby is only removed for the exact duplicated visible-label case, preserving composite labels.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
dandriscoll: Fixed by excluding aria-hidden labels from the duplicate-label lookup and adding preservation coverage.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
```diff
@@ -28,6 +28,23 @@ const AdaptiveCardAccessibilityWrapper: React.FC<{
...
+                const labelledBy = (select.getAttribute("aria-labelledby") || "").trim().split(/\s+/);
+                if (visibleLabel?.id && labelledBy.length === 1 && labelledBy[0] === visibleLabel.id) {
+                    select.removeAttribute("aria-labelledby");
+                }
```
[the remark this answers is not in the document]
dandriscoll: Fixed by requiring the associated label to be AT-visible before removing aria-labelledby.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
BJohnsonMicrosoft, line 27: nit: Consider renaming this function as it is now patching selects in addition to radio groups.

BJohnsonMicrosoft (approved):

VijVerma (approved):

ashishkadam0 (approved):

PR #927, opened 2026-05-08
fix(a11y): announce post-chat loading subtitle via role=status

dandriscoll:
> ⚠️ **Stacked on #918** — the diff currently includes the accessibility-tooling foundation commits. Once #918 merges this PR will rebase down to just the loading-pane status-region fix.

## Bug

When the post-chat survey is loading, `PostChatLoadingPaneStateful` renders a "Loading..." subtitle but does not place it inside a live region. NVDA / JAWS users get **no signal** that the survey is loading — the screen appears silent until the survey arrives.

## Fix

Add `role="status"` (implicit `aria-live="polite"`) to the loading-pane subtitle so screen readers announce the loading state.

## Tests

Un-skips the `loading-pane-status` catcher in `PostChatLoadingPaneStateful.a11y.spec.tsx` (added by #918).

---


ashishkadam0 (approved):

MugdhaSMundhe (approved):

chat-widget/src/components/postchatloadingpanestateful/PostChatLoadingPaneStateful.tsx
```diff
@@ -35,6 +35,7 @@ export const PostChatLoadingPaneStateful = (props: ILoadingPaneProps) => {
 
     const controlProps: ILoadingPaneControlProps = {
         id: "oc-lcw-postchatloading-pane",
+        role: "status",
```
jatinjd2904: Worth flagging: live regions that mount with text already inside them are not reliably announced by all screen readers — NVDA usually catches it, JAWS often misses, VoiceOver is inconsistent. The robust pattern is to render the region empty on mount and set the subtitle text on the next tick, or at minimum add `aria-atomic="true"` so the whole region is read on subsequent updates. Since the loading pane mounts with `subtitleText` already populated, the announcement may silently fail on the AT mix this PR is targeting. The unfixed-spec only checks the role is present, not that it actually speaks.

jatinjd2904 (approved):

chat-widget/src/components/postchatloadingpanestateful/PostChatLoadingPaneStateful.tsx
dandriscoll: Fixed: the live region now mounts empty and announces the subtitle after mount. Follow-up #934 tracks standardizing this test pattern.

chat-widget/src/components/postchatloadingpanestateful/PostChatLoadingPaneStateful.tsx
```diff
@@ -35,6 +35,7 @@ export const PostChatLoadingPaneStateful = (props: ILoadingPaneProps) => {
...
         dir: state.domainStates.globalDir,
         hideIcon: true,
```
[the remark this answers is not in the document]
dandriscoll: Fixed by enforcing the required status/live-region semantics after consumer props are spread. Follow-up #935 tracks a broader prop override audit.

chat-widget/src/components/postchatloadingpanestateful/PostChatLoadingPaneStateful.tsx
```diff
@@ -24,6 +33,14 @@ export const PostChatLoadingPaneStateful = (props: ILoadingPaneProps) => {
...
     }, []);
 
+    useEffect(() => {
+        const timeout = window.setTimeout(() => {
+            setAnnouncedSubtitleText(subtitleText);
+        }, 0);
+
+        return () => window.clearTimeout(timeout);
+    }, [subtitleText]);
+
```
BJohnsonMicrosoft, line 43: This causes a brief flash of empty content for sighted users. Between mount and the setTimeout firing, the subtitle briefly renders as an empty string, which may be perceptible on slower devices. It might be good to add a comment explaining that this behavior is intentional so a future maintainer doesn't try to "fix" it.

Example:
```
// Render the live region empty first, then populate subtitleText via setState.
// Screen readers only announce *mutations* inside an existing live region.
// If role="status" appears with text already inside, screen readers stay silent.
```

BJohnsonMicrosoft (approved):

VijVerma (approved):

VijVerma (approved):

VijVerma (approved):

Issue #919, opened 2026-05-08
a11y: classify catcher specs by confidence (high-confidence vs. best-effort proxy)


dandriscoll:
Follow-up from PR #918 review (suggestion by @deeptijoshim).

Today our a11y catcher specs don't visually distinguish between:

- **High-confidence catchers** that assert the DOM contract directly (e.g. `PostChatLoadingPaneStateful.a11y.spec.tsx` checking `role="status"`).
- **Best-effort proxy catchers** that test an indirect signal (e.g. `focusTrapAfterReload.spec.ts` checking `document.activeElement`).

A future contributor reading the suite has no fast way to tell which bugs are truly regression-guarded vs. which still need a manual screen-reader pass to confirm.

**Proposed:**

- Add a section to `docs/accessibility/README.md` defining the two confidence tiers with examples.
- Adopt a naming convention so it's visible at a glance — e.g. `*.a11y.proxy.spec.ts` for best-effort proxies, plain `*.a11y.spec.ts` for direct DOM contract assertions.
- One-time pass over existing catchers to rename / classify and link each to the bug it guards.

Out of scope for #918 because:

1. PR #918 is the foundation (tooling + skipped catchers); this is a polish pass over the catcher specs themselves.
2. There are 6 in-flight per-bug fix PRs queued behind #918; renaming spec files now would conflict with each of them on rebase.

Best done as a single sweep after #918 + the 6 fix PRs land.

PR #931, opened 2026-05-09
fix(vrt): stabilize postchatsurveypane VRT by intercepting iframe network

### LINKED ISSUE #921: VRT: postchatsurveypane stories embed live cross-origin iframe + fixed waitForTimeout, causing intermittent failures
## Summary

The `PostChatSurveyPane` Storybook stories embed a real cross-origin URL in an `<iframe>` (`https://ncv.microsoft.com/MPLJKJ3AlF`, a live Microsoft Forms survey) and the visual regression test waits a fixed `waitForTimeout: 5000` before snapshotting. Whether the iframe finishes loading inside 5 s depends on real network latency, so the baseline (captured when it had loaded) and a given test run (where it may still show "Loading…") can disagree by ~50 % of pixels. This produces an intermittent VRT failure on `postchatsurveypane-default-timeout-chromium` (and the analogous `preset-one` story).

The flake was surfaced and partially mitigated in PR #918 (accessibility-tooling foundation), but the underlying brittleness — real network fetch + fixed timer + no readiness signal — predates that PR and will resurface any time the test environment perturbs iframe load timing (CDP emulation calls, Chromium version bumps, ncv.microsoft.com latency spikes, etc.).

## Evidence

- Failing snapshot (CI run 25581976276 on PR #918, commit `f33d66a8`): three-panel diff shows baseline = fully-loaded survey form, actual = "Loading…" spinner, ~52.84 % differing pixels.
- Story source (`chat-components/src/components/postchatsurveypane/PostChatSurveyPane.stories.tsx` lines 23 and 34) hard-codes `surveyURL: "https://ncv.microsoft.com/MPLJKJ3AlF"`.
- Playwright actions (`PostChatSurveyPane.stories.playwright.json`): `[{ action: "waitForTimeout", options: 5000 }]` for both `Default` and `PresetOne` × `{chromium, firefox, webkit}`.

## Files involved

- `chat-components/src/components/postchatsurveypane/PostChatSurveyPane.stories.tsx`
- `chat-components/src/components/postchatsurveypane/PostChatSurveyPane.stories.playwright.json`
- (similar pattern may exist in other stories that embed external URLs — worth a sweep)

## Proposed approaches (any of, ordered by robustness)

1. **Stop hitting the network.** Replace the live `surveyURL` in the story with one of:
   - a `data:text/html,...` URL containing a static survey-shaped fixture, or
   - a local static HTML file shipped alongside the story (served by Storybook), or
   - a Storybook decorator that intercepts the iframe and renders a fixture in its place.
2. **Replace the fixed timeout with a deterministic wait.** Wait for an in-iframe DOM marker, a `load` event on the `<iframe>`, or for the "Loading…" element to detach. Combine with (1) so the wait is bounded by JS execution rather than network.
3. **Make the snapshot tolerant.** If the survey body is genuinely off-topic for what these stories are supposed to verify, mask the iframe region (CSS `visibility:hidden` overlay applied via decorator, or playwright `mask:` option on the snapshot call).
4. **Last resort: raise `failureThreshold`** for these stories specifically. Discouraged because it papers over future regressions in the survey-pane chrome (header/title/buttons).

Approach (1) + (2) together is the right long-term fix; (3) is acceptable if the embedded survey UI is not part of what the test is intended to verify.

## Acceptance criteria

- `postchatsurveypane-default-timeout-{chromium,firefox,webkit}` snapshots pass deterministically across 10 consecutive CI runs.
- No outbound request to `ncv.microsoft.com` (or any external survey provider) during VRT.
- Same robustness applied to any other story embedding an external URL (sweep + list, then fix).

## Related

- Surfaced in PR #918 review/CI cycle.
- Workaround in #918 commit `961fd3d2` removed an `emulateMedia` call from the `desktop-default` profile that was perturbing iframe load order; this only narrows the race window, it does not eliminate the underlying brittleness.
- Reviewer comments on #918 also flagged the pattern of fixed `waitForTimeout` + external URL.

dandriscoll:
Closes #921

## Problem

`PostChatSurveyPane` stories embed a live cross-origin iframe (`https://ncv.microsoft.com/MPLJKJ3AlF`) and the visual regression test waits a fixed `waitForTimeout: 5000` before snapshotting. Whether the iframe loads inside 5 s depends on real network latency, so the captured baseline (loaded) and a given run (still showing the browser's "Loading…" placeholder) can disagree by ~50 % of pixels. This produces intermittent failures on:

- `stateless-components-postchatsurvey-pane-post-chat-survey-pane-default-post-chat-survey-pane-default-timeout-{chromium,firefox,webkit}-snap.png`
- `stateless-components-postchatsurvey-pane-post-chat-survey-pane-preset-one-post-chat-survey-pane-preset-one-timeout-{chromium,firefox,webkit}-snap.png`

## Fix

Combines the issue's preferred approaches **(1) replace external content** and **(2) deterministic wait**:

1. **Network interception in test setup.** `chat-components/jest.setup.visual.js` now installs a Playwright `page.route()` handler inside `getPage()` that fulfills any request to known external survey hosts (`ncv.microsoft.com`, `customervoice.microsoft.com`, `tip.customervoice.microsoft.com`) with a small static fixture HTML. This satisfies the no-outbound-network requirement **without** changing component behavior or the story files, so `yarn storybook` (manual review) still hits the live survey.
   - Approach (1) as written in the issue (swap the story URL for `data:text/html,…`) was attempted first but is blocked by `getValidatedURL` in `chat-components/src/common/utils.ts`, whose regex only accepts `http(s)://`, `www.`, `mailto:`, `tel:`, `sms:`. A `data:` URL would be rewritten to `about:blank` by the component before reaching the iframe. Test-layer interception is a clean equivalent that needs no production code changes.
2. **Shorter deterministic wait.** Reduced `waitForTimeout` from `5000` → `1000` in `PostChatSurveyPane.stories.playwright.json` (all 6 occurrences). With the interceptor, the iframe content is byte-identical and synchronously available; 1000 ms is comfortable headroom and not on any iframe-load critical path.
3. **Regenerated baselines.** Six PNGs (3 browsers × 2 stories) updated to the fixture content. They render identically across chromium/firefox/webkit (verified by two consecutive green runs).

## Validation

- `yarn eslint jest.setup.visual.js src/components/postchatsurveypane` — clean (1 pre-existing warning in `PostChatSurveyPane.tsx` unrelated to this change).
- `yarn test:visual` scoped to `PostChatSurveyPane.stories.playwright.json`: **6 / 6 snapshots passed**, run twice (22 s each) → deterministic.
- Manual diff of regenerated baselines: identical fixture across browsers; no live-survey content leaked.
- Did not run the full VRT suite (per task brief — pre-existing flakes on `callingcontainer-windows-webkit` and `citationpane research-citation-init-webkit` are unrelated).

## Sweep — other stories embedding external URLs (not fixed here)

While diagnosing, found these for follow-up consideration; none are in the currently failing VRT, so I deliberately left them out of this PR's scope:

| File | External resource | Risk |
|---|---|---|
| `chat-widget/stories/component-stories/PostChatSurveyPane.stories.tsx` | `tip.customervoice.microsoft.com` (iframe) | Same class of flake if added to chat-widget VRT |
| `chat-widget/stories/OmniChannelLiveChatWidget.stories.tsx` | `purepng.com` / `freeiconspng.com` (CSS background images) | Lower — image, not iframe; smaller diff if it fails to load |
| `chat-widget/src/components/livechatwidget/LiveChatWidget.stories.tsx` | `pbs.twimg.com` (CSS background image) | Same as above |

The interceptor in this PR already covers the chat-widget survey hosts if those stories are ever pulled into a chat-components-style VRT, but the chat-widget package has its own test setup, so a follow-up should mirror this pattern there if/when the corresponding VRT is added.

## Out of scope

- File renaming for `*.a11y.proxy.spec.ts` convention (issue #919).
- Any changes to `getValidatedURL` itself — keeping the production allow-list intact is correct; the workaround belongs in test infra.


chat-components/jest.setup.visual.js
```diff
@@ -41,6 +41,46 @@ beforeAll(async () => {
...
+                + "input,textarea{width:100%;font-size:14px;padding:6px;box-sizing:border-box;border:1px solid #ccc}"
+                + "button{background:#0078d4;color:#fff;border:0;padding:8px 16px;font-size:14px;margin-top:12px}"
+                + "</style></head><body>"
+                + "<h1>Post-chat survey (test fixture)</h1>"
+                + "<div class=\"q\"><label>How would you rate your experience?</label>"
+                + "<input type=\"text\" value=\"\" readonly></div>"
+                + "<div class=\"q\"><label>Comments</label><textarea rows=\"3\" readonly></textarea></div>"
+                + "<button type=\"button\">Submit</button>"
+                + "</body></html>";
+            await page.route("**/*", async (route) => {
```
MugdhaSMundhe: The current glob pattern is too broad, as it now applies to every Storybook asset, JS bundle, font, and image, resulting in a new URL() allocation for each request. To optimize this, please scope the route specifically to the survey hosts. This adjustment will allow the handler to consistently fulfill requests, eliminating the need for try/catch blocks and unconditional route.continue().
  await page.route(
      /^https?:\/\/(ncv\.microsoft\.com|([a-z]+\.)?customervoice\.microsoft\.com)\//,
      (route) => route.fulfill({
          status: 200,
          contentType: "text/html; charset=utf-8",
          body: surveyFixtureHtml
      })
  );

MugdhaSMundhe (approved):

ashishkadam0 (approved):

chat-components/jest.setup.visual.js
dandriscoll: Thanks, good call. I updated this locally to scope the interceptor directly to the survey hosts with a route regex, so it no longer hooks every Storybook asset request or needs the URL allocation/try-catch/route.continue() path. Lint passes for the updated setup file.

dandriscoll: Addressed in prior commit.

jatinjd2904 (commented):
A few nits, all optional:

1. The `5000 → 1000` change is fine but it’s not really tied to the interceptor — `waitForTimeout` is just a hard sleep, so the iframe being faster doesn’t actually shorten it. Saves ~24s either way. Once the fixture has baked in for a release or two it’d be nice to swap this for `waitForLoadState("networkidle")` or a `waitForSelector` and drop the arbitrary wait.

2. Regex is a bit tight on subdomains — `([a-z]+\.)?customervoice.microsoft.com` only catches a single lowercase segment, so `tip2.`, `tip-test.`, or anything multi-level slips through. `ncv.microsoft.com` has no subdomain coverage. Doesn’t matter today, but `([a-z0-9-]+\.)*` would be more forward-proof.

3. Worth dropping a TODO in `jest.setup.visual.js` (or filing a follow-up) for the hosts in your sweep table — the chat-widget `PostChatSurveyPane` story and the `purepng.com` / `pbs.twimg.com` background images. Easy to lose track of those otherwise.

4. Tiny style thing — `surveyFixtureHtml` would read cleaner as a template literal vs. the `+ "..."` chain. Totally take-it-or-leave-it.

Nothing blocking, LGTM 👍

jatinjd2904 (approved):

VijVerma (approved):

ashishkadam0 (approved):

MugdhaSMundhe (approved):

PR #945, opened 2026-05-15
test(a11y): deterministic repro catchers for 5 remaining LCW a11y bugs

dandriscoll:
## Summary
Introduces 5 deterministic, failing-by-design accessibility repro catchers for the remaining LCW a11y bugs (catchers cover: ChatButton browse-mode duplicate stops, blank live-region announcements, AdaptiveCard duplicate labels on non-radio inputs, agent profile name announcement, and focus-trap leak across page reload).

## Status
All 5 catchers are intentionally wrapped in `describe.skip` on this foundation branch so CI stays green. Each accompanying source-fix PR un-skips its own catcher.

## Companion fix PRs
- #946 - ChatButton browse-mode
- #947 - WebChat toaster live region label
- #948 - AdaptiveCard non-radio duplicate labels
- #949 - Agent profile name e2e regression guard
- #950 - Focus trap leak across page reload

## Notes
- Catcher specs include diagnostic offender JSON output to `reports/` for debugging when run locally with the sample bundle built.
- Internal tracking IDs intentionally redacted.

PR #946, opened 2026-05-15
fix(a11y): consolidate ChatButton browse-mode announcement

dandriscoll:
## Summary
In screen-reader browse mode (NVDA/JAWS arrow-key navigation, Forms mode off) the ChatButton announces both the visible Title/Subtitle text AND the synthesized aria-label, producing duplicate stops in the reading order.

## Fix
- TextContainer outer Stack gets `aria-hidden` when no title/subtitle componentOverrides are present, so the visible text is hidden from the a11y tree.
- ChatButton synthesizes its aria-label from unread fragment + titleText + subtitleText so the button has a single, deterministic accessible name.
- Override-safe: when consumers supply title/subtitle componentOverrides, the gating skips so consumer-provided markup keeps its native a11y semantics.
- Falls through `controlProps?.ariaLabel` and `defaultChatButtonControlProps.ariaLabel` before applying the synthesized name.

## Verification
- Un-skipped catcher (renamed `.unfixed.a11y.spec.tsx` to `.a11y.spec.tsx`); 3/3 catcher tests pass.
- 13/13 broader chatbutton tests pass.
- `yarn build` clean.

## Related
- Catcher introduced in #945.
- Part of the a11y-bug fix series.

### DIFF
--- a/chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
+++ b/chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
@@ -1,0 +1,116 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import "@testing-library/jest-dom/extend-expect";
+
+import { cleanup, render } from "@testing-library/react";
+
+import { BroadcastServiceInitialize } from "../../services/BroadcastService";
+import ChatButton from "./ChatButton";
+import React from "react";
+import { defaultChatButtonProps } from "./common/defaultProps/defaultChatButtonProps";
+
+/**
+ * Repro / catcher for internal tracking — NVDA / JAWS in browse mode (virtual cursor)
+ * land on "Let's Chat, We're Online" multiple times when navigating with the
+ * down-arrow key.
+ *
+ * Root cause in code: `ChatButton` renders a Fluent UI `<Stack role="button">`
+ * with `tabIndex={0}` whose accessible name is computed from its visible
+ * children — a `<Label>` for the title (e.g. "Let's Chat!") and a `<Label>`
+ * for the subtitle (e.g. "We're online."). In browse mode, NVDA / JAWS treat
+ * the button itself AND each inner text node as separate virtual-cursor stops,
+ * each carrying the same announceable name fragments. There is no
+ * `aria-label` set by default, and the inner labels are not marked
+ * `aria-hidden="true"` (which is what would collapse them into a single
+ * announcement under the button container).
+ *
+ * Catcher contract: in the rendered chat-button subtree, at most ONE element
+ * may expose each of the title / subtitle strings as an announceable name
+ * source. An "announceable name source" here is:
+ *   - an `aria-label` attribute on a focusable / role-bearing element, OR
+ *   - a visible `<Label>` (or other text-bearing element) that is NOT
+ *     `aria-hidden="true"` and is not nested inside an element that is
+ *     `aria-hidden="true"`.
+ *
+ * Expected to FAIL today: the rendered DOM contains BOTH the title text node
+ * AND the subtitle text node as bare visible text, and the parent Stack has
+ * no `aria-label` to consolidate them. Browse-mode therefore sees three
+ * stops (button, title, subtitle) all carrying the same combined name.
+ */
+
+beforeAll(() => {
+    BroadcastServiceInitialize("testChannel");
+});
+
+afterEach(() => {
+    cleanup();
+});
+
+const isInsideAriaHidden = (el: Element | null): boolean => {
+    let cursor: Element | null = el;
+    while (cursor) {
+        if (cursor.getAttribute("aria-hidden") === "true") return true;
+        cursor = cursor.parentElement;
+    }
+    return false;
+};
+
+const visibleTextElementsMatching = (root: HTMLElement, text: string): Element[] => {
+    return Array.from(root.querySelectorAll("*")).filter((el) => {
+        if (isInsideAriaHidden(el)) return false;
+        // Only count leaf-ish text-bearing elements (don't double-count
+        // ancestors whose textContent obviously aggregates children).
+        const ownText = Array.from(el.childNodes)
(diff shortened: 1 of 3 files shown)


chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Catcher contract: in the rendered chat-button subtree, at most ONE element
+ * may expose each of the title / subtitle strings as an announceable name
+ * source. An "announceable name source" here is:
+ *   - an `aria-label` attribute on a focusable / role-bearing element, OR
+ *   - a visible `<Label>` (or other text-bearing element) that is NOT
+ *     `aria-hidden="true"` and is not nested inside an element that is
+ *     `aria-hidden="true"`.
+ *
+ * Expected to FAIL today: the rendered DOM contains BOTH the title text node
```
BJohnsonMicrosoft: This comment says "Expected to FAIL today" but a fix is applied in the same PR, so the test should be expected to pass now.

chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
BJohnsonMicrosoft, line 1: The three new cases all render `<ChatButton {...defaultChatButtonProps} />` (the default props path). None exercise:

- componentOverrides.title/subtitle/textContainer — which are the explicit "preserve consumer behavior" branches the fix advertises.
- A consumer-supplied `controlProps.ariaLabel` — which should win over the synthesized label.
- `hideNotificationBubble: false` with `unreadMessageCount: "3"` — which should produce the documented "3 you have new messages Let's Chat! We're online." label.

The fix's most complex logic (the gating) is exactly what's untested. Adding 2-3 cases here would meaningfully harden the regression guard.

BJohnsonMicrosoft (changes requested):

chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
dandriscoll: Addressed in d059bc87: added 4 new test cases covering (1) hideNotificationBubble=false + unreadMessageCount > 0 produces the unread-fragment label, (2) consumer-supplied controlProps.ariaLabel wins over synthesis, (3) componentOverrides.textContainer preserves consumer text and skips synthesis, (4) componentOverrides.title preserves consumer title and skips synthesis. All 7 tests pass locally.

chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
dandriscoll: Addressed in d059bc87: doc-comment header now reads 'Regression guard: ... a single consolidated aria-label ... preserved unchanged so customers who manage their own accessible names are not affected.'

BJohnsonMicrosoft (approved):

chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
dandriscoll: Addressed in d059bc87 (already on the branch from the previous review round). The doc comment now reads 'Regression guard: in the rendered DOM, the parent role=button Stack owns a single consolidated aria-label ...' - the stale 'Expected to FAIL today' line is gone.

chat-components/src/components/chatbutton/ChatButton.browseMode.a11y.spec.tsx
dandriscoll: Addressed in d059bc87 (already on the branch). Added 4 test cases covering exactly those gaps: (1) hideNotificationBubble=false + unreadMessageCount=3 producing the unread-fragment label, (2) consumer-supplied controlProps.ariaLabel winning over synthesis, (3) componentOverrides.textContainer preserving consumer text + skipping synthesis, (4) componentOverrides.title preserving consumer title + skipping synthesis. All 7 tests pass locally.

chat-components/src/components/chatbutton/ChatButton.tsx
```diff
@@ -100,14 +110,42 @@ function TextContainer(props: IChatButtonProps, parentId: string) {
...
+    const hideChatSubtitle = props.controlProps?.hideChatSubtitle ?? defaultChatButtonControlProps?.hideChatSubtitle;
+    const unreadMessageCount = props.controlProps?.unreadMessageCount ?? defaultChatButtonControlProps?.unreadMessageCount;
+    const ariaLabelUnreadMessageString = props.controlProps?.ariaLabelUnreadMessageString ?? defaultChatButtonControlProps?.ariaLabelUnreadMessageString;
+    const canSynthesizeAriaLabel = !hideChatTextContainer
+        && !props.componentOverrides?.textContainer
+        && !props.componentOverrides?.title
+        && !props.componentOverrides?.subtitle;
+    const synthesizedAriaLabel = canSynthesizeAriaLabel
+        ? [
+            !hideNotificationBubble && unreadMessageCount && unreadMessageCount !== "0"
```
jatinjd2904: `unreadMessageCount !== "0"` is a brittle string-equality check. An empty string slips through (`"".trim()` → falsy on the `unreadMessageCount &&` clause so this is OK), but `" 0 "`, `"0.0"`, or a numeric `0` coerced via `String(0)` all sneak past and produce labels like `"0 you have new messages Let’s Chat!"`. The unit suite only exercises `"3"` so this would land silently. Safer to coerce-and-check: `Number(unreadMessageCount) > 0`.

dandriscoll: Addressed in 8d5f6101: introduced a hasUnreadMessages(unreadMessageCount) helper that coerces via Number() and checks > 0, so ' 0 ', '0.0', '', and non-numeric noise no longer slip past. Added a parameterized regression block exercising each of those values for both the bubble rendering and the synthesized aria-label.

jatinjd2904 (approved):

PR #947, opened 2026-05-15
fix(a11y): label WebChat toaster live region and remove dead file-sent alert

dandriscoll:
## Summary
Screen readers announce empty live regions as `blank` when traversing the LCW page.

Two offenders fixed:

1. **BasicToaster (WebChat)** - renders `<div role="log">` with no accessible name. Added a mount-time effect that sets `aria-label` on `.webchat__toaster[role='log']` (with a short polling retry to handle late mount). Consumers may override via the new `webChatNotificationRegionAccessibilityLabel` prop (defaults to `Chat notifications`).
2. **Static `role=alert` div** (`fileSentAnnouncementRegionId`) in `WebChatContainerStateful` - dead code. `announceFileSent` dynamically creates fresh announcement nodes appended to `document.body`; the static element was never populated and screen readers walked past it as a blank live region. Removed it; the dynamic announce path is unchanged.

## Verification
- Un-skipped `automation_tests/e2e/areas/accessibility/blankAnnouncements.spec.ts` - now a passing regression guard when the sample bundle is built.
- `yarn lint` passes.
- All 161 `WebChatContainerStateful` unit tests pass.

## Related
- Builds on the deterministic catcher from #945.

### DIFF
--- a/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
+++ b/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
@@ -247,8 +247,8 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
         // Inject an aria-label so the region has a meaningful name; the
         // consumer can override via the new
         // `webChatNotificationRegionAccessibilityLabel` prop. The toaster
-        // may not be rendered yet when this effect first runs, so poll a
-        // few times after mount.
+        // may not be rendered yet when this effect first runs, so observe
+        // for it via a MutationObserver and apply the label once it appears.
         const toasterLabel =
             webChatContainerProps?.webChatNotificationRegionAccessibilityLabel
             ?? defaultWebChatContainerStatefulProps.webChatNotificationRegionAccessibilityLabel
@@ -261,14 +261,15 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
             }
             return false;
         };
+        let toasterObserver: MutationObserver | undefined;
         if (!labelToaster()) {
-            let attempts = 0;
-            const toasterInterval = window.setInterval(() => {
-                attempts++;
-                if (labelToaster() || attempts >= 20) {
-                    window.clearInterval(toasterInterval);
+            toasterObserver = new MutationObserver(() => {
+                if (labelToaster()) {
+                    toasterObserver?.disconnect();
+                    toasterObserver = undefined;
                 }
-            }, 250);
+            });
+            toasterObserver.observe(document.body, { childList: true, subtree: true });
         }
         dispatch({
             type: LiveChatWidgetActionType.SET_RENDERING_MIDDLEWARE_PROPS,
@@ -292,6 +293,10 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
                 }
             }
         }
+
+        return () => {
+            toasterObserver?.disconnect();
+        };
     }, []);
 
     useEffect(() => {
--- a/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
+++ b/chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
@@ -248,28 +248,38 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
         // consumer can override via the new
         // `webChatNotificationRegionAccessibilityLabel` prop. The toaster
         // may not be rendered yet when this effect first runs, so observe
-        // for it via a MutationObserver and apply the label once it appears.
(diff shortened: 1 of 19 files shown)


chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -240,6 +240,36 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
...
+        };
+        if (!labelToaster()) {
+            let attempts = 0;
+            const toasterInterval = window.setInterval(() => {
+                attempts++;
+                if (labelToaster() || attempts >= 20) {
+                    window.clearInterval(toasterInterval);
+                }
+            }, 250);
+        }
```
BJohnsonMicrosoft: The enclosing useEffect returns no cleanup. If the widget unmounts within 5s of mount, the interval keeps firing and the closure holds onto stale references. Worst case is bounded (≤20 fires, ≤5s), so impact is small, but it's a real leak in StrictMode double-invoke and in fast open/close UX. Track the handle and clear it from a returned cleanup.

Also consider preferring a MutationObserver on webChatRoot (one already exists in this file). Same trigger source, no need for polling.

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -505,26 +535,6 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
...
-                    width: "1px",
-                    height: "1px",
-                    padding: "0",
-                    margin: "-1px",
-                    overflow: "hidden",
-                    clip: "rect(0, 0, 0, 0)",
-                    whiteSpace: "nowrap",
-                    border: "0"
-                }}
-            />
```
BJohnsonMicrosoft, line 527: Cleanup Opportunity: This static element was removed, but [announceFileSent](https://github.com/microsoft/omnichannel-chat-widget/blob/main/chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/storemiddlewares/attachmentUploadValidatorMiddleware.ts#L19) still tries to look it up by ID. Functionally it's fine but the lookup is dead and the comments are now misleading, might as well clean them up.

BJohnsonMicrosoft (changes requested):

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
dandriscoll: Addressed in f4c0d0d1: removed the dead getElementById('ms_lcw_file_sent_announcement') lookup + stale comments, and removed the now-unused fileSentAnnouncementRegionId constant from HtmlIdNames.

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
dandriscoll: Addressed in f4c0d0d1: replaced the setInterval-based polling with a MutationObserver scoped to document.body that disconnects once the toaster is labeled or the effect tears down (returned from the useEffect cleanup).

chat-widget/src/components/webchatcontainerstateful/WebChatContainerStateful.tsx
```diff
@@ -240,6 +240,37 @@ export const WebChatContainerStateful = (props: ILiveChatWidgetProps) => {
...
+        };
+        let toasterObserver: MutationObserver | undefined;
+        if (!labelToaster()) {
+            toasterObserver = new MutationObserver(() => {
+                if (labelToaster()) {
+                    toasterObserver?.disconnect();
+                    toasterObserver = undefined;
+                }
+            });
+            toasterObserver.observe(document.body, { childList: true, subtree: true });
```
jatinjd2904: Observing `document.body` with `subtree: true` fires the callback on every DOM mutation anywhere on the host page — customer dashboards / SPAs do this constantly. Even though `labelToaster` itself is cheap, the per-mutation observer cost stacks up on heavy host pages and was a real complaint behind earlier widget perf bugs. Scope to the widget root (`#oc-lcw-container` or the outer Stack ref) so the observer only fires inside the widget subtree.

dandriscoll: Addressed in 437b7ae8: the observer now resolves the widget root (#ms_lcw_webchat_root) and observes that scope; if the root isn't mounted yet the initial lookup falls back to document, but every observed mutation re-resolves and tightens to the widget root as soon as it appears, so we stop firing on host-page mutations once the widget is in the DOM.

jatinjd2904 (approved):

BJohnsonMicrosoft (approved):

PR #948, opened 2026-05-15
fix(a11y): strip duplicate labels on non-radio AdaptiveCard inputs

dandriscoll:
## Summary
TalkBack announces non-radio AdaptiveCard inputs (Input.Text, Input.Date, Input.Number, Input.Toggle, multi-select Input.ChoiceSet) up to three times because the renderer emits a visible `<label for>` AND `aria-label` AND `aria-labelledby` all resolving to the same string. TalkBack walks each accessibility node independently and reads the duplicate name.

## Fix
`AdaptiveCardAccessibilityWrapper` now also patches non-radio inputs in `.ac-input-container`: when a visible `<label for=id>` already names the control, the redundant `aria-label` / `aria-labelledby` on the input are removed so only one accessible-name source remains.

## Verification
- Un-skipped catcher: `AdaptiveCardAccessibilityWrapper.a11y.spec.tsx` (renamed from `.unfixed.a11y.spec.tsx`).
- 13/13 catcher tests pass (5 new non-radio cases + 8 existing radio / compact-dropdown / action-button assertions).
- 19/19 existing `AdaptiveCardAccessibilityWrapper.test.tsx` tests still pass.
- `yarn lint` clean.

## Related
- Catcher introduced in #945.

### DIFF
--- a/chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
+++ b/chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
@@ -45,6 +45,44 @@ const AdaptiveCardAccessibilityWrapper: React.FC<{
                 }
             });
 
+            // internal tracking: non-radio inputs (Input.Text, Input.Date, Input.Number,
+            // Input.Toggle, Input.ChoiceSet multi-select checkbox) rendered by
+            // adaptivecards can have THREE accessible-name sources resolving to
+            // the same string: a visible <label for>, aria-label, AND
+            // aria-labelledby. TalkBack walks each accessibility node
+            // independently and reads the duplicated name. When a visible
+            // <label for> exists and names the control, strip the redundant
+            // aria-label / aria-labelledby on the input so only one announceable
+            // source remains.
+            const labelledInputs = container.querySelectorAll<HTMLInputElement>(
+                ".ac-input-container input.ac-input[id]:not([type='radio'])"
+            );
+            labelledInputs.forEach((input) => {
+                const id = input.id;
+                if (!id) return;
+                const visibleLabel = container.querySelector<HTMLLabelElement>(
+                    `label[for="${CSS.escape(id)}"]:not([aria-hidden='true'])`
+                );
+                if (!visibleLabel) return;
+                const labelText = (visibleLabel.textContent || "").trim();
+                if (!labelText) return;
+                const ariaLabel = input.getAttribute("aria-label");
+                if (ariaLabel && ariaLabel.trim() === labelText) {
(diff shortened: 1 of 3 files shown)


BJohnsonMicrosoft (approved):
Looks good overall. Just want to confirm the multi-select ChoiceSet checkbox case will be validated on a real TalkBack device. The DOM contract test doesn't replicate TalkBack's tendency to land on label text instead of the input. If TalkBack lands on the label here, the user might still hear "Updates" but miss "checkbox, not checked".

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/attachments/AdaptiveCardAccessibilityWrapper.tsx
```diff
@@ -45,6 +45,44 @@ const AdaptiveCardAccessibilityWrapper: React.FC<{
...
+            // Input.Toggle, Input.ChoiceSet multi-select checkbox) rendered by
+            // adaptivecards can have THREE accessible-name sources resolving to
+            // the same string: a visible <label for>, aria-label, AND
+            // aria-labelledby. TalkBack walks each accessibility node
+            // independently and reads the duplicated name. When a visible
+            // <label for> exists and names the control, strip the redundant
+            // aria-label / aria-labelledby on the input so only one announceable
+            // source remains.
+            const labelledInputs = container.querySelectorAll<HTMLInputElement>(
+                ".ac-input-container input.ac-input[id]:not([type='radio'])"
```
jatinjd2904: Selector matches `input.ac-input` only, so `Input.Text` with `isMultiline: true` is missed — AdaptiveCards renders that as `<textarea class="ac-input ac-textInput ac-multichoiceInput">`, not `<input>`. Multiline text inputs are emitted with the same visible `<label for>` + `aria-label` + `aria-labelledby` triple that this PR fixes for single-line, so TalkBack will still double-read them. The non-radio test suite added below also doesn’t cover the multiline case. Widening the selector to `'.ac-input-container :is(input, textarea).ac-input[id]:not([type=\'radio\'])'` (or splitting into two queries) closes the gap.

dandriscoll: Addressed in 5dd4b260: widened the selector to '.ac-input-container input.ac-input[id]:not([type=radio]), .ac-input-container textarea.ac-input[id]' and extended the comment to call out the isMultiline=true textarea case. Added a regression catcher that mirrors the Input.Text (isMultiline=true) DOM (textarea.ac-input.ac-textInput.ac-multichoiceInput with the visible label + aria-label + aria-labelledby triple) and asserts at most one announceable name source survives.

jatinjd2904 (approved):

jatinjd2904 (approved):

ashishkadam0 (approved):

PR #956, opened 2026-06-01
fix(createMarkdown): default env to {} so markdown-it reference rules don't crash (fixes widget-disappearing on citation messages)

MSSunnyNayak:
## Summary

The chat widget can **crash and disappear from the page** when an agent message contains markdown reference-style links (`[label][id]`) or reference definitions (`[id]: url "title"`). The error in the browser console is:

```
TypeError: Cannot read properties of undefined (reading 'references')
  at Array.link (markdown-it/lib/rules_inline/link.js)
  at ParserInline.tokenize / parse
  at Core.process
  at MarkdownIt.parse
```

This was surfaced after consuming `@microsoft/omnichannel-chat-sdk@1.11.9-main.a5570e5` (which pins `acs_webchat-chat-adapter` to `0.0.1-beta.8`), but the **root cause is a latent bug in this repo**, not in the SDK or the adapter.

## Root cause (verified by reproduction)

`chat-widget/src/components/livechatwidget/common/createMarkdown.ts` overrides `md.render` and `md.renderInline` and forwards the caller-supplied `env` parameter (which is `undefined` for all internal callers — `initWebChatComposer.ts:159-160` calls `render(text)` with no env) straight to `md.parse` / `md.parseInline`.

`markdown-it`'s `StateCore` constructor stores `this.env = env` without defaulting. When env is `undefined`, the standard **inline `link` rule** and the standard **block `reference` rule** both do:

```js
if (typeof state.env.references === 'undefined') { return false }
```

JavaScript evaluates the property access `state.env.references` BEFORE applying `typeof`. With `state.env === undefined`, the property access throws — verbatim the error users see.

Standard `MarkdownIt.prototype.render` / `renderInline` already default env to `{}` internally. Our overrides simply forgot to.

## Why upgrading the SDK exposes this

The SDK bump from `1.11.9-main.268277c` → `1.11.9-main.a5570e5` is a **single-line change**: pin `acs_webchat-chat-adapter` from `^0.0.1-beta.6` (which was resolving to a rogue `0.0.1-beta-1` due to a semver §11 prerelease ordering edge case where the dash form sorts before the dotted form) to exact `0.0.1-beta.8`. This vaults 7 functional adapter releases at once:

| Aspect | Old (`0.0.1-beta-1`) | New (`0.0.1-beta.8`) |
|---|---|---|
| Activity-text shape for bot citations | Pre-processed / partial | Raw markdown body containing `[1]: cite:...` definitions and `[N]` references |
| Streaming RTN | No | Yes |
| Adaptive-card responses | No | Yes |

The new adapter now hands the widget raw citation markdown that flows through `citationsMiddleware.ts` → `replaceCitations` and lands in the markdown pipeline still containing reference-style patterns. Those patterns hit the standard markdown-it rules → crash.

The SDK / adapter is functionally correct. The widget's markdown wrapper is unsafe.

## Side effect — widget disappears

The throw happens inside the activity-rendering middleware pipeline, not purely inside React render. It escapes the React error boundary of `WebChatContainerStateful`, the subtree unmounts, and the widget visibly disappears from the page.

## Fix

Two-line defensive default in both `md.render` and `md.renderInline`:

```ts
const safeEnv = env ?? {};
```

Then forwarded to `md.parse` / `md.parseInline` / `md.renderer.render` in place of the raw `env` parameter. This restores the same invariant standard `MarkdownIt.prototype.render` already provides.

## How verified

New describe block in `createMarkdown.test.ts`: `createMarkdown - reference-link env safety` with 4 tests:

1. Reference-style citation definition + inline reference → throws without fix, passes with fix
2. Inline reference-style link `[a][b]` via `render` → throws without fix, passes with fix
3. Inline reference-style link `[a][b]` via `renderInline` → throws without fix, passes with fix
4. `disableMarkdownMessageFormatting: true` variant → passes (sanity check that the disabled path is also safe)

Without the patch, tests 1-3 fail with a byte-identical stack to the customer report:

```
TypeError: Cannot read properties of undefined (reading 'references')
    at Array.link (node_modules/markdown-it/lib/rules_inline/link.js:99:26)
    at ParserInline.tokenize (node_modules/markdown-it/lib/parser_inline.js:135:22)
    at ParserInline.parse (node_modules/markdown-it/lib/parser_inline.js:163:8)
    at Array.inline (node_modules/markdown-it/lib/rules_core/inline.js:10:23)
    at Core.process (node_modules/markdown-it/lib/parser_core.js:51:13)
    at MarkdownIt.parse (node_modules/markdown-it/lib/index.js:524:13)
    at MarkdownIt.render (src/components/livechatwidget/common/createMarkdown.ts:203:42)
```

With the patch: **24/24 createMarkdown tests pass** (20 existing + 4 new).

Command: `npx jest -c jest.config.unit.cjs --env=jsdom --runInBand --testPathPattern createMarkdown`

## Scope

Surgical. Only `createMarkdown.ts` (4 lines added) and `createMarkdown.test.ts` (regression tests). No behavior change for callers that already pass a valid env.

## Defense-in-depth follow-ups (not in this PR)

1. Wrap `renderMarkdown` invocation in `initWebChatComposer.ts` with try/catch fallback to plain text — so any *future* markdown-pipeline crash never takes down rendering.
2. React error boundary around `WebChatContainerStateful` — prevents *any* uncaught throw in the webchat subtree from unmounting the widget.

Happy to file these separately if owners want them.

image


jatinjd2904 (approved):

jatinjd2904:
@microsoft-github-policy-service agree company="Microsoft"

MohammadSuraj (approved):

PR #957, opened 2026-06-03
fix(prechat): restore full-size pre-chat pane (Bug 6423684)

aidhage:
## Bug
[Bug 6423684](https://dynamicscrm.visualstudio.com/OneCRM/_workitems/edit/6423684) — `[5.4 PPE][Exploratory] Pre-chat UI renders in reduced/small size when pre-chat is enabled in LCW`

## Root cause
#929 (commit adf2da05) wrapped `<PreChatSurveyPane>` in a new `<div onFocusCapture>` to host the polite live region. The new wrapper has no width/height styles, so the inner pane''s `height: inherit; width: inherit` now resolves against `auto` and the pane collapses to its intrinsic content size.

**Before #929:**
image

```tsx
return (<PreChatSurveyPane controlProps={...} styleProps={...} />);
```

**After #929:**

image

```tsx
return (
  <div onFocusCapture={announceFocusedElement}>
    <div role="status" aria-live="polite" .../>
    <PreChatSurveyPane ... />
  </div>
);
```

The outer `<div>` defaults to `display:block; width/height:auto`, breaking the `inherit` sizing chain that the legacy CRM consumer relies on (parent `Stack` has no explicit dimensions; pre-#929 the pane inherited through Composer''s WebChat sizing).

Every other pane (Loading, OOOH, PostChat, etc.) uses explicit `width: 100%; height: 100%` — only PreChat used `inherit`, which is why this regressed only the pre-chat surface.

## Fix
```diff
- <div onFocusCapture={announceFocusedElement}>
+ <div onFocusCapture={announceFocusedElement} style={{ width: "100%", height: "100%" }}>
```

Minimal, surgical change to the wrapper added by #929. The inner `<div role="status" aria-live="polite">` is **untouched** so the a11y live-region behavior is unchanged.

## Validation
- Built `chat-widget` and the `samples/typescript-sample` harness, hosted to Azure Blob, validated against PPE org (`m-cc6e263c-4df0-f011-93da-000d3a32c972.preprod.omnichannelengagementhub.com`).
- A/B compared no-fix vs. fix bundles side-by-side. Pre-chat pane renders at configured `50% × 600px` (not collapsed) with the fix.
- a11y unit test added by #929 (`PreChatSurveyPaneStateful.a11y.spec.tsx`, `prechat-stale-live-region` catcher) still passes — the live region structure is unchanged.

## Risk
Very low. Single-line CSS prop on a wrapper introduced ~3 weeks ago; restores the sizing behavior every pre-#929 release shipped with.


jatinjd2904 (approved):

jatinjd2904 (approved):

chat-widget/src/components/prechatsurveypanestateful/PreChatSurveyPaneStateful.tsx
```diff
@@ -235,7 +235,7 @@ export const PreChatSurveyPaneStateful = (props: IPreChatSurveyPaneStatefulParam
     }, []);
     
     return (
-        <div onFocusCapture={announceFocusedElement}>
+        <div onFocusCapture={announceFocusedElement} style={{ width: "100%", height: "100%" }}>
```
deeptijoshim, line 238: Did we test this for a11y specially on 400% zoom if it renders properly?

ashishkadam0 (approved):

rahuljms (approved):

chat-widget/src/components/prechatsurveypanestateful/PreChatSurveyPaneStateful.tsx
aidhage: Yes — validated at 400% zoom (WCAG 1.4.10 Reflow, 320×600 viewport) by comparing this PR's build against `upstream/main` with this file reverted.

**Side-by-side metrics on the PreChatSurvey dialog:**

| Build | Dialog x | Width | Clipped left |
|---|---|---|---|
| PR #957 HEAD | -60px | 360px | 60px |
| `main` (this file reverted) | -60px | 360px | 60px |

**Identical.** Any left-edge clipping at <370px viewports is **pre-existing** at the outer shell/dialog level (`#oc-lcw` is fixed at 370px wide, anchored bottom-right). The wrapper `style={{width:"100%", height:"100%"}}` added in this PR is *inside* the dialog and cannot cause the dialog itself to overflow.

✅ Safe to merge from an a11y reflow perspective. Filing a separate bug for the shell positioning at viewports <370px.

aparnadubeyms (approved):

PR #167, opened 2022-11-23
css fixes

marghayk:
Fixing button width, border width of input field

**Before**:
before

**After**:
after

PR #958, opened 2026-06-04
fix(prechat): preserve ChoiceSet placeholder to keep required validation

coryeurom:
## Problem

In the pre-chat survey, a **required** compact `Input.ChoiceSet` fails validation when the user selects the **first** choice — submit is blocked. It only passes when a non-first option is chosen. Reproducible on desktop (not just iOS).

## Root cause

`PreChatSurveyPane` deletes the placeholder `<option>` from every compact ChoiceSet after render (an iOS blank-row workaround from PR #899), with **no platform guard** — it runs everywhere:

```js
const firstOption = select.options[0];
if (firstOption && firstOption.disabled && firstOption.hidden && firstOption.value === "") {
    select.removeChild(firstOption);
}
```

Adaptive Cards' compact ChoiceSet `value` getter (`card-elements.js`) is:

```js
return this._selectElement.selectedIndex > 0 ? this._selectElement.value : undefined;
```

It hard-assumes **index 0 is the placeholder**. Once the placeholder is deleted, the first real choice occupies index 0 → selecting it gives `selectedIndex === 0` → getter returns `undefined` → `isRequired` validation fails. Any non-first choice is `selectedIndex > 0` → passes.

## Fix

- Stop deleting the placeholder `<option>` — keep index 0 = placeholder so AC's validation works.
- Retain the iOS native-rendering override (`-webkit-appearance: none`, `height: auto`, `min-height`) and the `.ac-choiceSetInput-compact` CSS.
- Add a regression-guard unit test asserting the placeholder survives render on a required compact ChoiceSet (placeholder at index 0, first real choice at index 1).

## Testing

`PreChatSurveyPane.test.tsx` — 3/3 pass, including the new guard.

## iOS note

The blank-row workaround relied on deleting the placeholder; that deletion was the bug. The CSS override remains — the iOS blank-row should be re-verified on a device, and if it returns, handled non-destructively (e.g. placeholder prompt text) rather than by deletion.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


coryeurom:
Closing — this is NOT the regression cause. Rollback testing showed release_2605.2 (no LCW inline iosPrechatDropdownFix) is unaffected while release_2605.3 (with the inline fix, cherry-pick #167) is broken — both pin the same chat-widget 9284e84 which already contains this PR's #899 placeholder removal. So #899's synchronous, pre-attach placeholder removal is benign for validation AND is what fixes the iOS blank-row in the good build. Reverting it here would reopen ICM 51000000905069 without fixing the regression. The actual fix is removing the redundant LCW inline backport (CRM.OmniChannel.LiveChatWidget#179).

PR #967, opened 2026-07-01
fix: edited system messages not updating due to stale text cache (AB#6523665)

MugdhaSMundhe:
## Summary
Fixes a regression introduced in PR #895 where edited system messages (e.g. queue-position updates: `People ahead of you: 2` → `You're next in line.`) never re-rendered with the updated text.

## Root cause
Added an `originalSystemMessageTexts` `Map<string, string>` in `activityMiddleware.tsx` keyed by `activity.id` alone. On every subsequent render for the same activity id, the middleware returned the first-seen text — so edits to the same activity were invisible in the UI, even though `card.activity.text` was already updated by BotFramework/ACS.

## Fix
Remove the cache entirely and render `card.activity.text` on every pass.

## Why removing the cache is safe
The exponential `<br>` growth the cache was designed to prevent is no longer possible: PR #895 also refactored `handleSystemMessage` to stop mutating `card.activity.text` (it now writes to a local `const renderedHtml`). With no mutation, repeated `renderMarkdown` calls on the same pristine text are idempotent — the exact behavior the code had for years prior to #895.

## Scope
- Single file: `chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/activityMiddleware.tsx`
- 7 insertions / 11 deletions
- Preserves all other PR #895 changes (iOS CSS fixes, `defaultSystemMessageStyles` updates, `endChat.ts` wiring)
- `resetActivityMiddlewareCache()` still exported and still clears `loggedSystemMessages` + `lastRenderedAt`, so `endChat.ts` caller is unaffected

## Regression analysis
| Scenario | Verdict |
|---|---|
| Exponential `<br>` growth | ✅ Safe — `card.activity.text` never mutated |
| Re-renders on minimize/restore, scroll, focus | ✅ Safe — `renderMarkdown` is pure |
| Dedup via `nextVisibleActivity` | ✅ Unchanged |
| Telemetry (AverageWaitTime, QueuePosition) | ✅ Uses separate `loggedSystemMessages` array |
| Edited system messages | ✅ Now correctly re-render (intended fix) |

## Work item
AB#6523665 — https://dev.azure.com/dynamicscrm/OneCRM/_workitems/edit/6523665

## Testing
Verified locally with the typescript sample app connected to a live Omnichannel org: queue-position system messages now update in place as the queue advances.

Screenshot1
Screenshot2
Screenshot3
Screenshot4
Screenshot5
Screenshot6
Screenshot7


aparnadubeyms (approved):

MSSunnyNayak (commented):
Review — ✅ Approve (correct, well-reasoned fix)
Validated against source at HEAD (f129de2). This is the right fix for the queue-position regression (AB#6523665).

Correctness

Root cause is accurate. The id-only originalSystemMessageTexts cache froze the first-seen text; ACS delivers queue-position updates as edits that reuse the same activity id, so the cache-hit suppressed the new text.
"Safe to remove the cache" holds up. In handleSystemMessage, card.activity.text is now read-only — rendering writes to a local const renderedHtml and returns via dangerouslySetInnerHTML. The only card.activity.text = … assignment in the file is in the user-message path, guarded by isHtmlEncoded and self-flagged (idempotent) — it never touches system messages. So the historical exponential <br> growth (which came from the middleware writing rendered HTML back into card.activity.text before https://github.com/microsoft/omnichannel-chat-widget/pull/895) genuinely cannot recur, and renderMarkdown is pure → repeated renders are idempotent.
No collateral damage. resetActivityMiddlewareCache() still clears loggedSystemMessages + lastRenderedAt, so the endChat.ts caller is unaffected. Telemetry dedup (loggedSystemMessages) and nextVisibleActivity dedup are untouched.
Non-blocking suggestions

Add a regression test. activityMiddleware.spec.ts isn't touched and currently has no coverage for this scenario, so nothing prevents a re-regression. Suggest: render a system activity with id X and assert text A; then re-render the same id X with updated text + a new editedOn/version and assert the UI now shows B (not A). (Confirmed removing the cache breaks no existing test.)
Minor: file still ends without a trailing newline (pre-existing).
Logic is sound and the scope is tight (single behavioral file + changelog). Approving; would just ask for the regression test before merge so this class of bug can't silently return.

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/activityMiddleware.tsx
```diff
@@ -55,14 +51,15 @@ const handleSystemMessage = (next: any, args: any[], card: any, renderMarkdown:
...
 
-    const activityKey = card.activity.id || card.activity?.channelData?.clientmessageid;
-    if (activityKey && !originalSystemMessageTexts.has(activityKey)) {
-        originalSystemMessageTexts.set(activityKey, card.activity.text);
-    }
-    const sourceText = (activityKey && originalSystemMessageTexts.get(activityKey)) ?? card.activity.text;
+    // Render latest text on every pass. We deliberately do NOT cache the pre-render text
+    // (removed post-PR #895), because:
+    //  1. card.activity.text is no longer mutated (renderedHtml is a local variable), so
+    //     exponential <br> growth from repeated markdown rendering is not possible.
```
MSSunnyNayak: Review — ✅ Approve (correct, well-reasoned fix)
Validated against source at HEAD (f129de2). This is the right fix for the queue-position regression (AB#6523665).

Correctness

Root cause is accurate. The id-only originalSystemMessageTexts cache froze the first-seen text; ACS delivers queue-position updates as edits that reuse the same activity id, so the cache-hit suppressed the new text.
"Safe to remove the cache" holds up. In handleSystemMessage, card.activity.text is now read-only — rendering writes to a local const renderedHtml and returns via dangerouslySetInnerHTML. The only card.activity.text = … assignment in the file is in the user-message path, guarded by isHtmlEncoded and self-flagged (idempotent) — it never touches system messages. So the historical exponential <br> growth (which came from the middleware writing rendered HTML back into card.activity.text before https://github.com/microsoft/omnichannel-chat-widget/pull/895) genuinely cannot recur, and renderMarkdown is pure → repeated renders are idempotent.
No collateral damage. resetActivityMiddlewareCache() still clears loggedSystemMessages + lastRenderedAt, so the endChat.ts caller is unaffected. Telemetry dedup (loggedSystemMessages) and nextVisibleActivity dedup are untouched.
Non-blocking suggestions

Add a regression test. activityMiddleware.spec.ts isn't touched and currently has no coverage for this scenario, so nothing prevents a re-regression. Suggest: render a system activity with id X and assert text A; then re-render the same id X with updated text + a new editedOn/version and assert the UI now shows B (not A). (Confirmed removing the cache breaks no existing test.)
Minor: file still ends without a trailing newline (pre-existing).
Logic is sound and the scope is tight (single behavioral file + changelog). Approving; would just ask for the regression test before merge so this class of bug can't silently return.

MSSunnyNayak (approved):

chat-widget/src/components/webchatcontainerstateful/webchatcontroller/middlewares/renderingmiddlewares/activityMiddleware.tsx
MugdhaSMundhe: Done

ashishkadam0 (approved):

ashishkadam0 (approved):

PR #970, opened 2026-08-05
fix(chat-components): remediate dependencies for 1.2.0

bghcore:
## Version
`@microsoft/omnichannel-chat-components` **1.1.17 → 1.2.0**

## Bug 6562437 coverage
- upgrades Adaptive Cards to `3.0.6` and Swiper to `12.1.2`, the first clean compatible pairing for the critical Swiper prototype-pollution advisory
- updates Fluent UI v8 and `broadcast-channel` to patched runtime lines
- restores the `glob-parent` 5.1.2 dev-tool security resolution
- uses canonical public Yarn lock metadata with SHA-512 integrity

## Compatibility and cleanup
- preserves tree-shakeable Adaptive Cards ESM imports and remaps only the CommonJS build to the UMD entry
- validates ESM/CJS entry separation and packed Node 20 CommonJS imports
- adds iOS ChoiceSet index-0 regression coverage for Adaptive Cards 3
- adds BroadcastService v7 channel-disposal coverage
- makes Jest web-global polyfills conditional and restores diagnostic comments
- completes all workflow setup-node migration to `.nvmrc`
- documents Adaptive Cards/broadcast-channel major upgrades and reviewed cosmetic visual deltas

## Validation
- build and lint
- 160 unit tests passed (9 skipped)
- 19 accessibility tests passed (3 skipped)
- 231 visual snapshots passed across Chromium, Firefox, and WebKit
- packed consumer audit: 0 vulnerabilities
- ESM/CJS and packed CommonJS consumer validation passed

AB#6562437


chat-components/src/components/confirmationpane/ConfirmationPane.stories.tsx
```diff
@@ -277,9 +277,9 @@ export const ConfirmationPanePreset3: Story<IConfirmationPaneProps> = (args) =>
     args.controlProps.onConfirm = switchToRowLayout;
     args.controlProps.onCancel = switchToColumnLayout;
```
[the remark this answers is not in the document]
bghcore: Fixed by creating an immutable controlProps override only when args.controlProps exists. The story no longer mutates args and safely passes undefined through to the component.

chat-components/package.json
```diff
@@ -11,6 +11,9 @@
   "files": [
     "lib"
   ],
+  "engines": {
+    "node": ">=22.12.0"
+  },
```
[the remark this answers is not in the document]
bghcore: Removed the published package-level engines constraint. Node 22 remains enforced through .nvmrc and the component PR/release workflows without changing consumer install compatibility.

chat-components/package.json
```diff
@@ -69,22 +65,24 @@
...
+    "typescript": "^5.8.3",
+    "undici": "8.9.0"
   },
   "dependencies": {
-    "@fluentui/react": "^8.46.0",
-    "adaptivecards": "^2.10.0",
-    "broadcast-channel": "^4.5.0",
-    "rxjs": "^5.0.3"
+    "@fluentui/react": "^8.125.7",
+    "adaptivecards": "3.0.6",
```
ca1eb, line 74: **Blocking:** `adaptivecards@3.0.6` exposes ESM syntax through the entry loaded by `require("adaptivecards")`, while this package still advertises a CommonJS build through `lib/cjs/index.js`. Because `PreChatSurveyPane` imports Adaptive Cards and is eagerly re-exported from the root, CommonJS consumers can fail immediately with a syntax error when importing the package. Please retain a CommonJS-compatible Adaptive Cards version, bundle/transpile this dependency into the CJS artifact, or intentionally remove the CommonJS export as a breaking change.

ca1eb (changes requested):
The Adaptive Cards upgrade introduces a CommonJS compatibility regression that should be resolved before merging.

ca1eb (approved):
Same as the last - left one comment from the robot, ignore if it doesn't make sense

chat-components/package.json
bghcore: Reproduced and fixed. The component now imports the Adaptive Cards 3 UMD distribution, emits lib/cjs/package.json with type=commonjs, and CI validates the CommonJS export. A packed consumer import succeeds under Node 20.

ca1eb (approved):

bghcore:
The required build is pending because GitHub has not created any Actions runs for the current fork head (0 workflow runs/check suites; CLA is the only reported check). A repository maintainer needs to click **Approve and run workflows** for the latest commit. The component fix is locally validated, including packed CommonJS import on Node 20.

chat-components/src/common/utils.ts
```diff
@@ -1,6 +1,6 @@
 import { ElementType, Regex } from "./Constants";
 
-import { AdaptiveCard } from "adaptivecards";
+import { AdaptiveCard } from "adaptivecards/dist/adaptivecards.js";
```
coryeurom: **This fixes the CJS break, but it also applies to the ESM build — and that's a significant regression for LCW.**

I unpacked `adaptivecards@3.0.6` to compare the two entry points:

```
dist/adaptivecards.js       2765 KB   <- what this import now pulls in
dist/adaptivecards.min.js    327 KB
lib/  (the package `main`)   ~470 KB across ~15 split ES modules
```

`dist/adaptivecards.js` is a **pre-bundled, unminified UMD** blob. It's not tree-shakeable, and it has all of Swiper webpack-inlined into it:

```js
// dist/adaptivecards.js:8202
/* harmony import */ var swiper__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(/*! swiper */ "../node_modules/swiper/swiper.mjs");
// ...and the full swiper/modules/*.mjs graph inlined from L12726 onward
```

Whereas the package `main` (`lib/adaptivecards.js`) is a clean ESM barrel:

```js
// lib/adaptivecards.js
export * from "./strings";
```

Because `build:esm` and `build:cjs` both compile the same `src/`, this deep import lands in `lib/esm` as well. So LCW's webpack build — which resolves the `import` condition — now gets the 2.7 MB UMD instead of the tree-shakeable module graph, and loses the ability to drop `carousel.js` + Swiper (~25 KB + all of Swiper) that it almost certainly never renders.

There's a second consequence: `swiper@12.1.2` is added at line 77 to satisfy AdaptiveCards' `peerDependencies` (`{ "swiper": "^12.1.2" }` — I confirmed that's real and the direct dep is correct *in principle*). But on the UMD path Swiper is already inlined, so consumers install and ship Swiper **twice**: once standalone, once baked into the UMD.

**Suggested fix: scope the remap to the CJS build only.** Keep the bare specifier in source so `lib/esm` stays clean:

```ts
import { AdaptiveCard } from "adaptivecards";
```

…and alias it in `babel.config.json` (used exclusively by `build:cjs`), leaving `babel.esm.config.json` untouched:

```json
"plugins": [
  ["module-resolver", { "alias": { "^adaptivecards$": "adaptivecards/dist/adaptivecards.js" } }]
]
```

That keeps `test:cjs` green, keeps ESM consumers tree-shakeable and Swiper-deduped, and lets you drop `src/types/adaptivecards-dist.d.ts` entirely.

If you'd rather keep the single import for simplicity, at minimum please point it at `dist/adaptivecards.min.js` (327 KB vs 2765 KB) and add a comment explaining why the deep import exists, so nobody "fixes" it back to the bare specifier later and silently breaks CJS again.

chat-components/src/components/prechatsurveypane/PreChatSurveyPane.tsx
```diff
@@ -1,4 +1,4 @@
-import * as AdaptiveCards from "adaptivecards";
+import * as AdaptiveCards from "adaptivecards/dist/adaptivecards.js";
```
coryeurom: Same deep-import concern as `src/common/utils.ts` — see that thread for the bundle-size analysis and the suggested CJS-only aliasing approach. Whatever you land there should apply here too.

Separately, worth double-checking against this file's **iOS ChoiceSet workaround** (the one in the changelog that strips the placeholder `<option>` AdaptiveCards injects and patches the `ChoiceSetInput` value getter so `selectedIndex === 0` is accepted). That fix reaches into AdaptiveCards internals, so a 2 → 3 major bump is exactly the kind of change that can silently invalidate it — if AC 3 changed how `Input.ChoiceSet` renders in `compact` style, the patch could become either a no-op or actively wrong.

The VRT snapshots in this PR won't catch it, since the bug was iOS-Safari-specific. Could you confirm this was manually verified on iOS, or note it as a follow-up test?

chat-components/package.json
```diff
@@ -99,20 +98,13 @@
...
-    "**/ssri": "8.0.1",
-    "**/ansi-regex": "5.0.1",
-    "**/nanoid": "3.1.31",
-    "**/url-parse": "1.5.9",
-    "**/markdown-it": "12.3.2",
-    "**/minimist": "1.2.6",
-    "**/sanitize-html": "2.12.1",
-    "**/lodash": "4.17.23",
-    "**/brace-expansion": "2.0.2"
+    "**/@types/glob": "7.2.0",
```
coryeurom, line 106: **Dropping `**/glob-parent` reintroduces a known ReDoS advisory.**

On `main` the resolution collapsed every range onto the patched version:

```
# main
glob-parent@5.1.2, glob-parent@^3.1.0, glob-parent@^5.1.2, glob-parent@~5.1.2, glob-parent@^6.0.2:
  version "5.1.2"
```

Without it, the `^3.1.0` range floats back down:

```
# this PR
glob-parent@^3.1.0:   version "3.1.0"    <- GHSA-ww39-953v-wcq6, fixed in 5.1.2
glob-parent@^5.1.2:   version "5.1.2"
glob-parent@^6.0.2:   version "6.0.2"
```

It comes in via `chokidar@2.1.8` and `fast-glob@2.2.7` (the Storybook/webpack watch chain), so it is **dev-only** — which is consistent with the PR's "0 vulnerabilities" claim, since that was a *packed consumer* audit and this never reaches the shipped tree.

But on a PR whose stated purpose is dependency remediation, silently regressing a pinned advisory is the wrong direction, and a plain `yarn audit` will now flag it. Please either restore just this one entry:

```json
"**/glob-parent": "5.1.2",
```

…or, if it's being left deliberately because it's dev-only and unreachable, say so explicitly in `CHANGE_LOG.md` alongside the other removals. Right now the changelog lists the removed Markdown/sanitizer/Lodash/Nano ID/brace-expansion resolutions but doesn't mention `glob-parent`, `ssri`, `url-parse`, `minimist`, or `ansi-regex` at all.

For the record, I checked all the others and they're genuinely fine — `lodash` → 4.18.1 (higher than the old 4.17.23 pin), `nanoid` → 3.3.17, `url-parse` → 1.5.10, `minimist` → 1.2.8, `ssri` → 6.0.2/8.0.1, `brace-expansion` → 1.1.18/2.1.4/5.0.9, and `sanitize-html`/`markdown-it` dropped out of the tree entirely with the AdaptiveCards 2.x removal. `glob-parent` is the only regression.

coryeurom (commented):
Reviewed with a focus on what actually ships to consumers (chat-widget / LCW), since `adaptivecards` 2 → 3 and `broadcast-channel` 4 → 7 are both major bumps behind a minor version.

Good call on catching the CJS/ESM break and adding `test:cjs` to CI — that's exactly the right guardrail, and I verified the `lib/cjs/package.json` marker is genuinely required here (the package has `"type": "module"` at the root, so without it every `.js` under `lib/cjs` would be parsed as ESM). Nice fix.

Two findings I think are worth acting on before merge, plus some smaller ones — details inline.

**Summary**

| # | Item | Severity |
|---|---|---|
| 1 | Deep-importing the Adaptive Cards **UMD** bundle pushes a 2.7 MB non-tree-shakeable blob into `lib/esm` too, not just `lib/cjs` | Major |
| 2 | Dropping `**/glob-parent` regresses it 5.1.2 → 3.1.0, reintroducing GHSA-ww39-953v-wcq6 | Major (dev-only) |
| 3 | AdaptiveCards major bump + 7 changed screenshots shipped as a minor version | Discuss |
| 4 | `undici` polyfill overwrites Node 22's native `fetch`/`Blob`/streams unconditionally | Minor |
| 5 | `@babel/helper-split-export-declaration` pinned as a direct dep (removed from Babel in 7.25+) | Minor |
| 6 | `.nvmrc` migration is incomplete — same file mixes both styles | Nit |
| 7 | Unexplained comment deletions in `jest.setup.visual.cjs`, incl. the link to #921 | Nit |

I also verified the other removed `resolutions` are genuinely safe — `lodash` → 4.18.1, `nanoid` → 3.3.17, `url-parse` → 1.5.10, `minimist` → 1.2.8, `ssri` → 6.0.2/8.0.1, and `sanitize-html`/`markdown-it` are gone from the tree entirely (they came in via AdaptiveCards 2.x). Only `glob-parent` regressed. And confirmed the removed ESLint plugins are unreferenced by `eslint.config.js`, so that cleanup is clean.

One cross-PR note: microsoft/omnichannel-chat-sdk#577 adds a package-level `engines.node` constraint, while this PR explicitly avoids one. I think this PR made the right call — but the two should agree.

chat-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@microsoft/omnichannel-chat-components",
-  "version": "1.1.17",
+  "version": "1.2.0",
```
coryeurom, line 3: Worth an explicit decision: should this be **2.0.0** rather than 1.2.0?

The runtime dependency changes include two majors on packages whose behaviour is part of this package's observable surface:

- `adaptivecards` `^2.10.0` → `3.0.6` (two majors) — drives all pre-chat survey rendering
- `broadcast-channel` `^4.5.0` → `^7.3.0` (three majors) — backs `BroadcastService`, which is core public API

The PR itself is the strongest evidence that this is user-visible: it updates **7 VRT snapshots** across `prechatsurveypane` and `reconnectchatpane`. Those are, by definition, rendering changes that consumers will see. A consumer on `^1.1.x` running `yarn upgrade` would pick this up and get a different-looking pre-chat survey with no signal that anything changed.

Either bump to 2.0.0, or — if the team is confident the visual deltas are cosmetic and acceptable — please add a prominent `### Changed` note in `CHANGE_LOG.md` that (a) names AdaptiveCards 3 explicitly as a major upgrade and (b) summarises what visibly changed in the affected panes. Right now the changelog frames this purely as security + tooling, which undersells it.

On `broadcast-channel` specifically: `BroadcastService` uses `new BroadcastChannel(name)` and assigns `subChannel.onmessage = ...`. Both still work in v7, and I confirmed the `exports` map keeps a browser-safe entry (`browser` → `dist/esbrowser/index.js`) and a CJS entry (`main` → `dist/es5node/index.js`), so the dual-format story holds. But `close()` semantics and the internal method-selection defaults did shift across v5-v7 — is `BroadcastService` covered by the 158 unit tests, particularly the cross-tab teardown path?

chat-components/jest.setup.visual.cjs
```diff
@@ -1,9 +1,16 @@
...
-import storybookAccessibilityTooling from "../tools/accessibility/storybookProfiles.cjs";
+const { setConfig } = require("storybook-addon-playwright/configs");
+const { toMatchScreenshots } = require("storybook-addon-playwright");
+const storybookAccessibilityTooling = require("../tools/accessibility/storybookProfiles.cjs");
+
+const { ReadableStream, TransformStream, WritableStream } = require("node:stream/web");
+const { Blob, File } = require("node:buffer");
+const { MessageChannel, MessagePort } = require("node:worker_threads");
+Object.assign(globalThis, { Blob, File, MessageChannel, MessagePort, ReadableStream, TransformStream, WritableStream });
+const { fetch, FormData, Headers, Request, Response } = require("undici");
```
coryeurom, line 28: This polyfill block is at best redundant and at worst actively harmful on the Node version this PR standardizes on.

On Node 22, every one of these is **already a global**: `fetch`, `FormData`, `Headers`, `Request`, `Response`, `Blob`, `File`, `ReadableStream`, `TransformStream`, `WritableStream`, `MessageChannel`, `MessagePort`. The unconditional `Object.assign(globalThis, ...)` therefore doesn't fill a gap — it *replaces* Node's built-in undici with a different undici major (`8.9.0`).

That's a real risk for Playwright-driven VRT: mixing two undici realms means `instanceof Response` / `instanceof Blob` checks can fail depending on which side constructed the object, and those failures tend to surface as flaky, hard-to-diagnose screenshot timeouts rather than clean errors.

If these were added to fix a specific failure, please make them conditional so they're inert on Node 22:

```js
if (typeof globalThis.fetch !== "function") {
    const { fetch, FormData, Headers, Request, Response } = require("undici");
    Object.assign(globalThis, { fetch, FormData, Headers, Request, Response });
}
```

…with a one-line comment naming the environment that needs it. Otherwise I'd drop lines 7-12 and the `undici` devDependency (line 70) entirely — given the repo now pins Node 22 via `.nvmrc`, the natives should be sufficient.

chat-components/package.json
```diff
@@ -22,6 +22,7 @@
   "devDependencies": {
     "@babel/cli": "^7.26.10",
     "@babel/core": "^7.26.10",
+    "@babel/helper-split-export-declaration": "7.24.7",
```
coryeurom, line 25: Could you add a comment or PR note explaining this pin? `@babel/helper-split-export-declaration` is an **internal** Babel helper that was removed from Babel in 7.25+ — it isn't something a project normally depends on directly.

Pinning it at `7.24.7` as a direct devDependency looks like it's papering over a transitive break (most likely Storybook 6.5's Babel chain expecting a helper that newer `@babel/traverse` no longer ships). That works, but it's invisible to the next person: it'll look like a stray dependency and get "cleaned up", breaking the build in a way that's genuinely hard to trace back.

Two suggestions:
1. Express it as a `resolutions` entry rather than a direct devDependency — that's semantically what it is (constraining someone else's transitive dep), and it won't read as a first-party dependency.
2. Either way, note in `CHANGE_LOG.md` *what* needs it, so it can be removed when Storybook is eventually upgraded past 6.5.

Related: this PR migrates Storybook 6.5 to the Webpack 5 builder but stays on 6.5, which is EOL. Not something to fix here, but if there's no tracking item for the Storybook 7/8 migration, this pin is a good argument for opening one.

.github/workflows/npm-release.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 22.x
+          node-version-file: .nvmrc
```
coryeurom, line 38: Nit: the `.nvmrc` migration is partial, and this file ends up inconsistent with itself — line 38 uses `node-version-file: .nvmrc` but line 86 still hardcodes `node-version: 22.x`.

Across the repo, 4 of 12 `node-version` references were migrated:

```
chat-components-pr.yml:21            node-version-file: .nvmrc   ✅
chat-components-release.yml:20       node-version-file: .nvmrc   ✅
chat-components-release-manual.yml   node-version-file: .nvmrc   ✅
npm-release.yml:38                   node-version-file: .nvmrc   ✅
npm-release.yml:86                   node-version: 22.x          ❌ same file
chat-widget-pr.yml:29                node-version: 22.x          ❌
chat-widget-release.yml:24           node-version: 22.x          ❌
chat-widget-release-manual.yml:15    node-version: 22.x          ❌
storybook-release.yml:27             node-version: 22.x          ❌
accessibility-scan.yml:104           node-version: 22.x          ❌
accessibility-sr.yml:47              node-version: 22.x          ❌
```

Since the whole point of `.nvmrc` is a single source of truth, a half-migration arguably leaves things worse than before — there are now two mechanisms to keep in sync instead of one.

Either finish the migration (it's a mechanical change and `22.x` already floats to whatever `.nvmrc` pins, so it's low-risk), or at minimum fix line 86 in this file. If the intent was to scope this PR strictly to `chat-components`, that's reasonable — just say so in the description, and please still fix line 86 since it's in a file you're already touching.

(FWIW I confirmed `22.23.2` is a real, current Node 22 LTS release — published 2026-07-28. Good pin.)

chat-components/jest.setup.visual.cjs
```diff
@@ -1,9 +1,16 @@
...
+Object.assign(globalThis, { fetch, FormData, Headers, Request, Response });
+const playwright = require("playwright");
 
 expect.extend({ toMatchScreenshots });
```
coryeurom, line 36: Nit — this file loses several explanatory comments that don't seem related to the dependency remediation, including:

- the rationale for intercepting external survey hosts, **and its link to microsoft/omnichannel-chat-widget#921**
- the note explaining why `HOOK_TIMEOUT_MS` is 120s while the per-test timeout is 50s (slow CI agents launching 3 browsers)
- the diagnostic comment explaining why leftover contexts/pages are logged before `browser.close()` — i.e. what to look at if the close ever hangs again

Those all encode debugging history that took someone real effort to establish. Could you restore them? The issue link in particular is the only breadcrumb back to the original investigation.

Also in this file: `preparePageForProfile(page, screenshotProfile)` moved out of page creation into a new `beforeScreenshot` hook. That's a genuine behavioural change to when viewport/UA emulation is applied relative to navigation — plausibly the reason some snapshots changed. Worth calling out in the PR description rather than leaving it to be inferred from the diff, since it makes the snapshot churn much easier to reason about.

The `clearTimeout(timeoutId)` addition in the teardown race is a nice catch, though — that was a real handle leak keeping the process alive.

chat-components/src/common/utils.ts
bghcore: Implemented CJS-only remapping with a Babel plugin used only by babel.config.json. ESM keeps the bare adaptivecards import and test:cjs now asserts ESM remains tree-shakeable while CJS uses the UMD entry.

chat-components/src/components/prechatsurveypane/PreChatSurveyPane.tsx
bghcore: Added a deterministic iOS ChoiceSet regression test. It verifies the placeholder removal, first visible value selection, and Adaptive Cards 3 ChoiceSetInput getter returning index 0.

chat-components/package.json
bghcore: Restored the glob-parent 5.1.2 resolution. The regenerated public lock no longer contains glob-parent 3.1.0.

chat-components/package.json
bghcore: Kept 1.2.0 because public APIs remain compatible, but expanded the Changed changelog to name both major dependency upgrades and the reviewed cosmetic baseline deltas. Added BroadcastService dispose coverage confirming both v7 channels close.

chat-components/jest.setup.visual.cjs
bghcore: Made all Jest web-global assignments conditional. Web Streams are installed before requiring undici (required by Jest 27), and Node 22 globals are never replaced when already present.

chat-components/package.json
bghcore: Documented that the direct 7.24.7 helper pin is a Storybook 6.5 compatibility constraint and should be removed with the tracked Storybook 7/8 migration. A direct pin is required here because Storybook loads the helper by package name.

.github/workflows/npm-release.yml
bghcore: Completed the migration: all workflow setup-node steps now use .nvmrc, including the second npm-release job and accessibility jobs.

chat-components/jest.setup.visual.cjs
bghcore: Restored the survey issue #921, slow-CI hook timeout, and teardown diagnostic comments. Also documented that profile emulation now runs after Storybook readiness to avoid iframe-load races.

coryeurom (approved):

PR #971, opened 2026-08-07
fix!: clean up chat-widget dependencies for 2.0.0 and retain required WebChat hotfix

bghcore:
## Version
`@microsoft/omnichannel-chat-widget` **1.8.4 → 2.0.0**

## Breaking change
Node.js `>=22.12.0` is now the supported consumer build runtime. This requirement matches chat SDK 2.0.0.

Consumers on Node 20 must upgrade Node before they use this release.

## Bug 6562437 package cleanup
- Consumes chat SDK `2.0.0-main.b82e941` and chat-components `1.2.0-main.5692126`.
- Retains required `botframework-webchat@4.18.1-hotfix.20260308.b15b405`.
- Upgrades Markdown-It 14.3.0, DOMPurify 3.4.x, and `sanitize-html` 2.17.6.
- Pins legacy RxJS `5.5.12` because the in-package adapters still use RxJS 5 deep imports. Transcript download intentionally continues to load its separate RxJS 7 browser global.
- Keeps Redux 5 as a development-only typecheck dependency; the published middleware declaration is structural and self-contained.
- Removes unused ES5 forks that carried stale build tools.
- Removes obsolete security resolutions.
- Uses canonical public Yarn lockfiles with SHA-512 integrity and no Microsoft-feed URLs.
- Scopes packed-consumer audit exceptions by package, installed version, advisory, and reviewed dependency roots (WebChat, plus chat SDK for the shared Azure Signaling UUID path).

## Package cleanup and compatibility
- Removes unused DOMPurify types, legacy ESLint packages, file-loader, p-defer, and the React Docgen TypeScript plugin.
- Migrates package, sample, UMD, and Storybook builds to Webpack 5.
- Migrates middleware to the Redux 5 `unknown` action contract without rejecting payload-less actions.
- Preserves the Storybook Playwright screenshot-authoring panel in development; only static Webpack 5 builds exclude the incompatible preset/register integration.
- Preserves and checks the browser-targeted CommonJS export.
- Deduplicates Trusted Types and pins compatible Storybook and Jest types.
- Moves Husky prepare logic into a linted script and prevents hook setup in external package consumers.
- Makes the CommonJS metadata writer independent of the caller's working directory.
- Verifies both the declared and installed WebChat hotfix versions from the README constraint.
- Fails the packed-consumer workflow immediately if packing, installation, or package presence validation fails.

## Validation
- Node.js 22 lint passes.
- 711 unit tests pass, including payload-less action and Markdown-It 14 plugin-chain regressions.
- ESM and CommonJS compilation pass.
- Audit validator fixtures confirm reviewed WebChat-owned findings pass, dangling `via` references warn with their parent, and a first-party dependency regression using the same advisory is rejected.
- Full GitHub validation is rerunning on the pushed review commit.

## Required WebChat hotfix exceptions
- `markdown-it@13.0.2` retains two ReDoS advisories.
- `linkify-it@4.0.1` retains two quadratic-complexity DoS advisories.
- `@babel/runtime@7.14.8`, `7.15.4`, and `7.19.0`, plus `@babel/runtime-corejs3@7.20.13`, retain the named-capture-group ReDoS advisory.
- `adaptivecards@3.0.2` retains vulnerable `swiper@8.4.7`; the Swiper advisory covers prototype pollution.
- `valibot@1.1.0` retains the emoji ReDoS and inherited-key error-flattening advisories.
- `sanitize-html@2.14.0` retains its advisory. The WebChat allowlist excludes the vulnerable `action` and `formaction` attributes.
- UUID 3, 8, and 9 retain the buffer-bounds advisory.
- The reviewed WebChat and Azure Signaling paths do not pass a caller-provided UUID buffer.
- A new WebChat hotfix must preserve the product fix and update these dependencies.

AB#6562437


bghcore:
Dependency remediation and package cleanup are complete. The widget workflow passed all gates, including unit, E2E, Storybook, 87 VRT snapshots, browser-targeted CommonJS bundling, and packed-consumer advisory validation. The sibling chat-components check is being rerun after a transient registry.yarnpkg.com core-js timeout.

bghcore:
Packed-consumer audit gate passed. It fails on any new advisory and permits only three reviewed upstream roots: GHSA-vccv-cmxp-4j9h (WebChat sanitize-html allowlist excludes affected attributes), GHSA-w5hq-g745-h8pq (UUID buffer APIs are unreachable in the used paths), and GHSA-5qjj-4xww-7phc (WebChat pins 1.4.1 for external npm consumers; first-party/Yarn builds are forced to patched Valibot 1.4.2).

chat-widget/src/components/livechatwidget/common/createMarkdown.ts
```diff
@@ -1,8 +1,8 @@
 import { Constants } from "../../../common/Constants";
 import MarkdownIt from "markdown-it";
 import MarkdownItForInline from "markdown-it-for-inline";
-import StateCore from "markdown-it/lib/rules_core/state_core";
-import Token from "markdown-it/lib/token";
+import StateCore from "markdown-it/lib/rules_core/state_core.mjs";
+import Token from "markdown-it/lib/token.mjs";
```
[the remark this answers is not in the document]
bghcore: Fixed in 676a53ba. Both Markdown-It internals are now type-only imports, so Babel erases them and the CommonJS build no longer emits `require()` calls for the `.mjs` modules.

chat-widget/src/components/webchatcontainerstateful/interfaces/IWebChatAction.ts
```diff
@@ -1,5 +1,24 @@
...
+
+/**
+ * Type guard narrowing the `unknown` action received by a Redux 5 store middleware
+ * down to Web Chat's IWebChatAction shape, so middlewares can safely access `type`/`payload`.
+ */
+export const isWebChatAction = (action: unknown): action is IWebChatAction =>
+    typeof action === "object" &&
+    action !== null &&
+    "type" in action &&
+    "payload" in action;
```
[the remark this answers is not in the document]
bghcore: Fixed in 676a53ba. The guard now requires a string `type` and an own `payload` property before narrowing the Redux 5 `unknown` action.

chat-widget/scripts/validate-consumer-audit.cjs
```diff
@@ -0,0 +1,55 @@
+const { readFileSync } = require("node:fs");
+
+const report = JSON.parse(readFileSync(process.argv[2], "utf8"));
+const vulnerabilities = report.vulnerabilities ?? {};
```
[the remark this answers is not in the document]
bghcore: Fixed in 676a53ba. The validator now fails closed when npm audit returns a top-level error or when `vulnerabilities` is missing, null, or not an object. I exercised error, missing, array, and valid-empty report shapes locally.

chat-widget/scripts/validate-consumer-audit.cjs
```diff
@@ -0,0 +1,55 @@
...
+const allowedAdvisories = new Set([
+    "https://github.com/advisories/GHSA-5qjj-4xww-7phc",
```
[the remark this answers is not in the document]
bghcore: Investigated against the public registry and upstream source. `botframework-webchat@4.19.1` is still the latest published release and pins `valibot@1.4.1`; unpublished WebChat `main` also still pins 1.4.1, so there is no patched WebChat release to consume. A widget-side Yarn resolution cannot protect external npm consumers, and adding another override would repeat the ad hoc behavior this PR removes. We are retaining the explicit reviewed exception until WebChat publishes the fixed dependency; the advisory requires WebChat to validate attacker-controlled records and call `flatten()`, which is not a reachable WebChat path in this package.

chat-widget/src/components/webchatcontainerstateful/interfaces/IWebChatAction.ts
```diff
@@ -1,5 +1,25 @@
...
+/**
+ * Type guard narrowing the `unknown` action received by a Redux 5 store middleware
+ * down to Web Chat's IWebChatAction shape, so middlewares can safely access `type`/`payload`.
+ */
+export const isWebChatAction = (action: unknown): action is IWebChatAction =>
+    typeof action === "object" &&
+    action !== null &&
+    "type" in action &&
+    typeof action.type === "string" &&
+    Object.prototype.hasOwnProperty.call(action, "payload");
```
coryeurom: **The `payload` own-property check makes this guard reject legitimate store actions.**

Web Chat dispatches plenty of payload-less actions, and this repo does too — `localizedStringsBotInitialsMiddleware` dispatches `dispatch({ type: "__BOT_INITIALS_UPDATED__" })` with no `payload` (twice: in `resetCachedAgentName` and in the initials-changed branch). Those actions now fail `isWebChatAction` and hit the `return next(action)` short-circuit that this PR added to the top of all 16 migrated middlewares.

Today that's benign, because every guarded branch keys off action types that always carry a payload. But it's a latent trap with an invisible failure mode: a future middleware that needs to observe a payload-less action will simply never fire — no error, no log, no test failure.

The check also buys nothing. Every call site already uses `action.payload?.…`, and `IWebChatAction.payload` is `any`. Suggest dropping it:

```ts
export const isWebChatAction = (action: unknown): action is IWebChatAction =>
    typeof action === "object" &&
    action !== null &&
    "type" in action &&
    typeof action.type === "string";
```

chat-widget/package.json
```diff
@@ -70,46 +67,44 @@
...
+    "dompurify": "^3.4.12",
+    "markdown-it": "^14.3.0",
     "markdown-it-attrs": "^4.1.6",
-    "markdown-it-attrs-es5": "^2.0.2",
     "markdown-it-for-inline": "^0.1.1",
     "md5-typescript": "^1.0.5",
-    "p-defer-es5": "^2.0.1",
-    "sanitize-html": "2.14.0",
+    "redux": "^5.0.1",
+    "rxjs": "^5.5.12",
```
coryeurom: Declaring `rxjs` is the right call — `mockadapter.ts` is reachable from `defaultWebChatContainerStatefulProps.ts`, so it was a phantom dependency before this PR. Two concerns with the version:

1. **`^5.5.12` is EOL.** RxJS 5's last release was 2018. Adding a new EOL runtime dependency in a PR whose stated purpose is dependency remediation deserves an explicit callout — the description currently says "Upgrades … Redux 5, and RxJS", which reads like RxJS went *up*.
2. **The caret is unsafe here.** `mockadapter.ts`, `DemoChatAdapter.ts`, `DesignerChatAdapter.ts`, and `utils/chatAdapterUtils.ts` all use RxJS 5-only deep paths (`rxjs/Observable`, `rxjs/Subscriber`, `rxjs/add/operator/share`). `^5.5.12` is safe today only because 5.x is dead. Given this PR pins `sanitize-html`, `undici`, and `storybook-addon-playwright` exactly, pinning `5.5.12` would be consistent and self-documenting.

Also worth reconciling while you're here: `defaultLibraryScripts.ts` loads **rxjs 7.8.0** from cdnjs at runtime and `createChatTranscript.ts` reads `window.rxjs.BehaviorSubject`. So the widget will bundle RxJS 5 *and* load RxJS 7 from a CDN. Intentional?

chat-widget/package.json
```diff
@@ -70,46 +67,44 @@
...
-    "markdown-it": "^12.3.2",
+    "dompurify": "^3.4.12",
+    "markdown-it": "^14.3.0",
     "markdown-it-attrs": "^4.1.6",
-    "markdown-it-attrs-es5": "^2.0.2",
     "markdown-it-for-inline": "^0.1.1",
     "md5-typescript": "^1.0.5",
-    "p-defer-es5": "^2.0.1",
-    "sanitize-html": "2.14.0",
+    "redux": "^5.0.1",
```
coryeurom: `redux` is used only for `import type { Middleware }` in `IWebChatAction.ts` — nothing needs it at runtime. It's presumably in `dependencies` because the emitted `lib/types/**/*.d.ts` references the type, but that ships a second Redux into every consumer's tree alongside the Redux 4 that `botframework-webchat@4.18.1-hotfix.*` already brings.

Two things worth confirming:

- **Which Redux typings does `createStore` resolve against in `initWebChatComposer.ts`?** Redux 4 types middleware `next` as `Dispatch`; Redux 5 types it as `(action: unknown) => unknown`. Under `strictFunctionTypes` those aren't freely interchangeable, so it'd be good to know the compat is real rather than incidental to how WebChat's `.d.ts` happens to be written.
- **Could this be a `devDependency`?** Declaring the middleware shape structurally in `IWebChatAction.ts` (it's ~4 lines) would keep the public `.d.ts` self-contained and avoid shipping a runtime package that's never required.

chat-widget/package.json
```diff
@@ -70,46 +67,44 @@
...
+    "@microsoft/omnichannel-chat-sdk": "2.0.0-main.b82e941",
     "@opentelemetry/api": "^1.9.0",
     "abort-controller": "^3",
-    "abort-controller-es5": "^2.0.1",
     "botframework-webchat": "4.18.1-hotfix.20260308.b15b405",
     "core-js-pure": "^3.42.0",
-    "dompurify": "^3.2.4",
-    "markdown-it": "^12.3.2",
+    "dompurify": "^3.4.12",
+    "markdown-it": "^14.3.0",
```
coryeurom, line 102: `markdown-it` 12 → 14 is two majors on the message-rendering path, and neither the description nor the changelog mentions re-validating the plugin chain. v14 swaps in `linkify-it@5` and reworks entity handling, and `createMarkdown.ts` runs with `html: true` + `linkify: true`.

Were `slack-markdown-it@^1.0.5`, `markdown-it-for-inline@^0.1.1`, and `markdown-it-attrs@^4.1.6` re-verified against v14? `markdown-it-for-inline` hasn't been published since 2016 and this PR leaves all three ranges untouched.

Concretely at risk in `createMarkdown.ts`:
- `markdown.disable(["strikethrough"])` and the `markdown.enable(["entity", "linkify", "html_block", "html_inline", "newline", "list"])` list — markdown-it throws on unknown rule names, so a renamed rule is a hard runtime failure in the `disableMarkdownMessageFormatting` path.
- The `merge_adjacent_same_href_links` core rule reads `token.attrs` as `string[][]` and mutates `blockToken.children` — both are internal-ish shapes that moved between majors.

DOMPurify runs downstream so this isn't a sanitization hole, but a silent rendering regression here is directly user-visible. A unit test over the Slack-markdown + link-merge paths would be worth more than the audit gate here.

chat-widget/package.json
```diff
@@ -151,7 +148,7 @@
     "prepush": "yarn lint && yarn danger:prepush",
     "danger:prepush": "yarn danger local --dangerfile dangerfile.js --base main",
     "pretest:visual": "yarn playwright install",
-    "prepare": "cd .. && husky chat-widget/.husky",
+    "prepare": "node -e \"const fs=require('node:fs'),cp=require('node:child_process'),path=require('node:path');const repoRoot=path.resolve('..');if(fs.existsSync(path.join(repoRoot,'.git'))){cp.execFileSync(process.execPath,[path.resolve('node_modules','husky','bin.js'),'chat-widget/.husky'],{cwd:repoRoot,stdio:'inherit'});}\"",
```
coryeurom: This inline `node -e` blob is hard to read, impossible to lint, and awkward to change — and this PR already adds a `scripts/` folder with three `.cjs` files. Suggest `"prepare": "node ./scripts/prepare.cjs"` with the same logic in a real file, which also gets ESLint coverage and a place to explain *why* the `.git` probe exists.

Minor correctness note: `path.resolve('node_modules', 'husky', 'bin.js')` resolves against `chat-widget/`, but the child process runs with `cwd: repoRoot`. It works because the path is made absolute first, but that's subtle enough to deserve a comment — or just use `require.resolve('husky/bin.js')`.

chat-widget/scripts/write-cjs-package.cjs
```diff
@@ -0,0 +1,3 @@
+const { writeFileSync } = require("node:fs");
+
+writeFileSync("lib/cjs/package.json", "{\"type\":\"commonjs\"}\n");
```
coryeurom: This writes to a **cwd-relative** path, unlike its sibling `validate-webchat-version.cjs`, which correctly uses `resolve(__dirname, "..")`. It only works because `build:cjs` happens to run from `chat-widget/`; invoked from the repo root it either throws `ENOENT` or silently writes to the wrong `lib/cjs/`.

```js
const { writeFileSync } = require("node:fs");
const { resolve } = require("node:path");

writeFileSync(resolve(__dirname, "..", "lib", "cjs", "package.json"), "{\"type\":\"commonjs\"}\n");
```

chat-widget/scripts/validate-webchat-version.cjs
```diff
@@ -0,0 +1,14 @@
+const { readFileSync } = require("node:fs");
+const { resolve } = require("node:path");
+
+const requiredVersion = "4.18.1-hotfix.20260308.b15b405";
+const packageJson = JSON.parse(readFileSync(resolve(__dirname, "..", "package.json"), "utf8"));
+const configuredVersion = packageJson.dependencies?.["botframework-webchat"];
```
coryeurom, line 9: This validates the *declared* range in `package.json`, not what actually got installed. A `resolutions` entry, a stale `yarn.lock`, or a hoisting change could still put a different `botframework-webchat` in `node_modules` and this gate would pass — which defeats the purpose, since the thing being protected is the hotfix *code*, not the string in `package.json`.

The workflow runs this right after `yarn install`, so checking the resolved version is nearly free and much stronger:

```js
const installedPath = resolve(__dirname, "..", "node_modules", "botframework-webchat", "package.json");
const installedVersion = JSON.parse(readFileSync(installedPath, "utf8")).version;
```

…and assert both. (Reading it via `require("botframework-webchat/package.json")` may be blocked by the package's `exports` map, hence the explicit path.)

Separately: `requiredVersion` is now hardcoded in three places — here, `package.json`, and the README table. A future hotfix bump that misses this file fails with a message telling the author the *new* version is wrong.

chat-widget/scripts/validate-consumer-audit.cjs
```diff
@@ -0,0 +1,72 @@
...
+    typeof report.vulnerabilities !== "object" ||
+    report.vulnerabilities === null ||
+    Array.isArray(report.vulnerabilities)
+) {
+    console.error("Packed-consumer audit did not return a valid vulnerability report.");
+    process.exit(1);
+}
+
+const vulnerabilities = report.vulnerabilities;
+const allowedAdvisories = new Set([
```
coryeurom: The allowlist is keyed on advisory URL alone, with no package or version scoping. That makes this gate meaningfully weaker than the PR description implies ("Rejects each new packed-consumer advisory").

Concretely: `markdown-it`, `sanitize-html`, and `uuid` all appear here as *nested* WebChat-hotfix copies, but the widget also declares `markdown-it@^14.3.0` and `sanitize-html@2.17.6` as **direct** dependencies. If a direct dependency later regresses into one of these same GHSAs, `npm audit` emits the identical URL and this script waves it through as a reviewed hotfix exception.

Since the stated invariant is "only the WebChat hotfix owns these", consider keying on the audit entry's `name` + `range` (e.g. `sanitize-html@2.14.0 -> GHSA-…`) so a first-party regression can't reuse a third-party exception.

Also: the ten GHSA IDs are opaque as written. The README table and PR body map them to packages — a one-line `//` comment per entry would keep that mapping from drifting the next time the hotfix moves.

chat-widget/scripts/validate-consumer-audit.cjs
```diff
@@ -0,0 +1,72 @@
...
+
+const collectAdvisories = (name, visited = new Set()) => {
+    if (visited.has(name)) {
+        return;
+    }
+    visited.add(name);
+
+    const vulnerability = vulnerabilities[name];
+    if (!vulnerability) {
+        missingReferences.add(name);
```
coryeurom: This turns an unresolvable `via` string into a hard CI failure. `npm audit --omit=dev` prunes dev-only entries from `vulnerabilities`, but `via` chains can still name a pruned package, so this can fail the PR gate for a reason unrelated to the widget's production dependencies — and `unresolved dependency reference: foo` gives the author no path to a fix.

Suggest either warning instead of failing, or at minimum printing the parent entry that produced the dangling reference so it's diagnosable from the CI log alone.

.github/workflows/chat-widget-pr.yml
```diff
@@ -55,6 +58,27 @@ jobs:
...
+        shell: pwsh
+        run: |
+          $consumer = Join-Path $env:RUNNER_TEMP "chat-widget-consumer"
+          $tarball = Join-Path $env:RUNNER_TEMP "chat-widget.tgz"
+          New-Item -ItemType Directory -Path $consumer -Force | Out-Null
+          yarn pack --filename $tarball --silent
+          Set-Content -Path (Join-Path $consumer "package.json") -Value '{"name":"chat-widget-consumer","version":"1.0.0","private":true}'
+          Push-Location $consumer
+          try {
+            npm install $tarball --ignore-scripts --no-fund
```
coryeurom, line 76: Under `shell: pwsh`, GitHub Actions only propagates the exit code of the **last** native command in the block — here that's `node …/validate-consumer-audit.cjs` on line 78. That's deliberate for `npm audit` (which exits non-zero whenever findings exist), but it also silently swallows failures from `yarn pack` (line 71) and `npm install` here.

That's the bad case: if the pack produces an empty/broken tarball or the install fails, `npm audit` reports zero vulnerabilities, the validator prints nothing and exits 0, and the gate goes green — precisely the scenario this step exists to catch.

```powershell
yarn pack --filename $tarball --silent
if ($LASTEXITCODE -ne 0) { throw "yarn pack failed" }
...
npm install $tarball --ignore-scripts --no-fund
if ($LASTEXITCODE -ne 0) { throw "npm install failed" }
```

Worth also asserting `node_modules/@microsoft/omnichannel-chat-widget` exists in the consumer dir before auditing, so an empty tree can never pass.

coryeurom (commented):
## Review summary

Reviewed the full diff (47 files) — the Webpack 5 migration, the Redux 5 middleware refactor, the three new validation scripts, and the dependency deltas. No blocking defect found, but there are a few things I'd want addressed before this ships as a major.

I verified the two things that looked riskiest and they're **fine** — flagging so nobody re-litigates them:

- `import type … from "markdown-it/lib/token.mjs"` / `state_core.mjs` resolves correctly: `@types/markdown-it@14.1.2` ships `lib/token.d.mts` and `lib/rules_core/state_core.d.mts`, and TS maps `.mjs` → `.d.mts`. Switching these to `import type` was also necessary, since markdown-it 14's `exports` map no longer exposes those subpaths at runtime.
- `jest.setup.visual.js` → `.cjs` is a rename (61% similarity), not an orphaned file.
- The deleted `botInitials.spec.ts` case really is redundant — the adjacent "bot messages use descriptive alt text" test waits on `.webchat__basic-transcript`, polls `[role='group']` for rendered activities, and asserts on their accessible names. Strictly stronger.
- `channelDataMiddleware`'s `action` → `updatedAction` rewrite and `htmlTextMiddleware` / `preProcessingMiddleware`'s `currentAction` locals are behavior-preserving.

### Main points (details inline)

1. **`isWebChatAction` requires an own `payload` property.** Payload-less actions — including ones this repo dispatches itself — now fail the guard and short-circuit through all 16 migrated middlewares. Benign today, invisible failure mode tomorrow. Easiest and safest fix in the PR.
2. **`rxjs@^5.5.12` is a new EOL runtime dependency** in a dependency-remediation PR, and the body's "Upgrades … RxJS" phrasing hides that. Declaring it is correct (it was a phantom dep via `mockadapter.ts`); the version and the framing need a callout.
3. **`validate-webchat-version.cjs` checks the declaration, not the resolution.** It runs immediately after `yarn install`, so asserting the installed version too is nearly free and much stronger.
4. **The consumer-audit gate has two soft spots**: the allowlist isn't scoped to a package, so a *first-party* `markdown-it`/`sanitize-html`/`uuid` regression can reuse a WebChat exception; and `pwsh` swallows `yarn pack` / `npm install` failures, so a broken tarball audits clean and passes.
5. **`markdown-it` 12 → 14 is two majors on the message-rendering path** with all three plugin ranges left untouched. Would like confirmation the Slack-markdown and link-merge paths were re-validated.
6. **Changelog is under `[Unreleased]` for a versioned 2.0.0 release**, unlike the two sibling PRs.
7. **`samples/typescript-sample` still pins the widget at `^1.8.2-main.*`.**

### Minor / non-blocking

- `webpack.config.cjs` and `webpack.dev.config.cjs` set both `resolve.alias["react-native"] = false` **and** `new webpack.IgnorePlugin({ resourceRegExp: /^react-native$/ })`. The alias alone is sufficient; the pair is harmless but confusing.
- `test:cjs` only *bundles* `lib/cjs/index.js` through Webpack — it never executes it. "Browser-targeted CommonJS import passes" in the description is a stronger claim than the check supports. A `node -e "require('./lib/cjs')"` smoke test would close that gap cheaply.
- The `@typespec/ts-http-runtime` `moduleNameMapper` block is now copy-pasted verbatim into all three Jest configs. Worth hoisting into a shared `jest.moduleNameMapper.cjs`.
- Dropping `react-docgen-typescript-plugin` for `typescript: { reactDocgen: "react-docgen" }` degrades Storybook prop tables (no TS type info). Fine tradeoff, but it's a docs regression that isn't mentioned.
- `**/sanitize-html: 2.14.0` is removed while the direct dep goes to `2.17.6`, so the tree now carries two copies (2.17.6 top-level, 2.14.0 under WebChat). Intentional per the description, just noting the bundle-size effect.

### On the relationship to omnichannel-amsclient#80 and omnichannel-sdk#196

These are a **different workstream** from this PR. #80 and #196 are release-automation changes (create a GitHub Release on `v*` tags, attach the packed npm tarball, reject prerelease/mismatched tags). This PR is dependency remediation plus a 2.0.0 major. They touch no common code.

The real connections are three:

1. **Consumption.** This PR takes `@microsoft/omnichannel-chat-sdk@2.0.0-main.b82e941`; #196 finalizes `@microsoft/ocsdk@0.6.0`, which chat-sdk sits on top of. Per the bottom-up release order (ocsdk → amsclient → chat-sdk → chat-widget → LCW), #196 and #80 should land and tag *before* chat-widget 2.0.0 is published, otherwise 2.0.0 ships pointing at prerelease `-main.<sha>` builds of its transitive stack.
2. **Node 22 floor.** All three raise the supported runtime to `>=22.12.0`. That's consistent, but it means the LCW host app's build agents need Node 22 before *any* of them can be consumed. Worth confirming that's already true.
3. **Changelog convention.** #80 and #196 both finalize a dated version heading; this PR doesn't (point 6 above).

One thing worth flagging on **#196 / #80** themselves, since it affects this repo as a consumer: both explicitly document that pushes to `main` publish prereleases to the npm **`latest`** dist-tag ("The workflow intentionally moves `latest` to each `main` prerelease. An install without a version can receive a prerelease."). That's a real hazard for anyone running `npm install @microsoft/ocsdk` unpinned. This repo is insulated because it pins exact `-main.<sha>` versions — but the `main` dist-tag existed for exactly this reason, and I'd push back on retiring it rather than documenting around it. Best raised on those PRs.

Also on #196: `actions/checkout@v4` is left unpinned in the new `github-release` job while `download-artifact` and `action-gh-release` are correctly SHA-pinned. Worth making consistent.

chat-widget/.storybook/main.cjs
```diff
@@ -9,17 +9,26 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     '@storybook/addon-a11y',
-    '@storybook/addon-knobs',
-    'storybook-addon-playwright/preset',
-    'storybook-addon-playwright/register'
+    '@storybook/addon-knobs'
```
coryeurom: `storybook-addon-playwright/preset` and `/register` are dropped from both Storybook configs, but the package stays in `devDependencies` (now pinned to `4.9.2`) and `jest.setup.visual.cjs` still imports `setConfig` and `toMatchScreenshots` from it.

The description covers the Webpack 5 migration but doesn't mention losing the addon. Dropping `/register` removes the screenshot panel from the Storybook UI, which is how these snapshots are normally authored and updated. Is this an intentional scope reduction, or a Webpack 5 incompatibility workaround?

Either way it belongs in the description — "87 visual snapshots pass" only proves the headless assertion path still works, not the authoring workflow. Same change is in `stories/.storybook/main.cjs`.

CHANGE_LOG.md
```diff
@@ -6,13 +6,29 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Breaking
```
coryeurom, line 11: `package.json` moves to `2.0.0`, but everything here lands under `## [Unreleased]`.

Both sibling release PRs — omnichannel-sdk#196 and omnichannel-amsclient#80 — finalize a dated `## [X.Y.Z]` heading and leave an empty `## [Unreleased]` above it. This should follow the same convention, otherwise the 2.0.0 release has no changelog section to point at (and if this repo ever adopts the same tag-driven GitHub Release automation, the notes-extraction step would find nothing).

chat-widget/samples/typescript-sample/package.json
```diff
@@ -1,6 +1,9 @@
     {
       "name": "my-sample-widget",
       "version": "0.0.1",
+      "engines": {
+        "node": ">=22.12.0"
+      },
       "dependencies": {
         "@fluentui/react": "^8.123.1",
         "@microsoft/omnichannel-chat-widget": "^1.8.2-main.d82a79b",
```
coryeurom: `engines.node >= 22.12.0` is added here, but the sample still depends on `@microsoft/omnichannel-chat-widget@^1.8.2-main.d82a79b` — a 1.x range that can never resolve to the 2.0.0 this PR produces. The sample now advertises the new Node floor while pulling the old widget.

Should be bumped to `^2.0.0` (or the matching `2.0.0-main.*` prerelease) so it actually exercises the release it ships alongside.

coryeurom (approved):

chat-widget/src/components/webchatcontainerstateful/interfaces/IWebChatAction.ts
bghcore: Agreed. I removed the `payload` own-property requirement, made `payload` optional, and added a regression test for the payload-less `__BOT_INITIALS_UPDATED__` action. The guard now narrows on a string `type` only.

chat-widget/package.json
bghcore: Agreed. RxJS is now pinned exactly to `5.5.12`, and the PR description/changelog explicitly call out that this is an intentional legacy dependency for the RxJS 5 deep-import adapters, not an upgrade. The RxJS 7 CDN global is a separate transcript-download implementation and is now documented as such.

chat-widget/package.json
bghcore: Confirmed and changed. `redux` is now a devDependency only, and `IWebChatAction.ts` declares the small middleware shape structurally so the emitted declarations do not reference Redux. The hotfix WebChat `createStore` declaration accepts middleware as `any[]`; our internal structural type retains the Redux 5 `unknown` action contract without shipping a second Redux runtime.

chat-widget/package.json
bghcore: Added focused regression coverage for the Markdown-It 14 path: Slack bold/italic syntax, adjacent same-href link merging, and the minimal formatting rule set (`entity`, `linkify`, HTML, newline, and list). These tests run through the actual `createMarkdown` plugin chain.

chat-widget/package.json
bghcore: Moved the prepare logic to `scripts/prepare.cjs`, added the reason for the `.git` probe, and resolved Husky from its exported entry before locating `bin.js` (the `husky/bin.js` subpath is not exported). The script is covered by the package lint.

chat-widget/scripts/write-cjs-package.cjs
bghcore: Fixed. The writer now resolves `lib/cjs/package.json` from `__dirname`, and I verified it after a CommonJS build when invoked outside `chat-widget/`.

chat-widget/scripts/validate-webchat-version.cjs
bghcore: Strengthened the guard to compare both the declared dependency and the installed `node_modules/botframework-webchat` version. The required version is now read from the README constraint, so the script no longer carries a third hardcoded copy.

chat-widget/scripts/validate-consumer-audit.cjs
bghcore: Implemented package/version/advisory scoping and dependency ownership checks. A finding is allowed only when its exact installed package/version is reachable exclusively through the required WebChat hotfix; the same advisory reached through any other widget dependency is rejected. Each allowlist entry now includes its package mapping and rationale.

chat-widget/scripts/validate-consumer-audit.cjs
bghcore: Changed dangling `via` references to warnings and included the parent relationship (for example, `markdown-it -> pruned-parent`) so the CI log is actionable without turning an npm audit pruning detail into a gate failure.

.github/workflows/chat-widget-pr.yml
bghcore: Added explicit `$LASTEXITCODE` checks after `yarn pack` and `npm install`, plus an assertion that the packed widget exists in the consumer tree before audit validation. Pack/install failures can no longer be masked by the final Node command.

chat-widget/.storybook/main.cjs
bghcore: The authoring workflow is preserved. Both Storybook configs now include the Playwright preset/register integration in development and exclude it only for static Webpack 5 builds via `STORYBOOK_BUILD_MODE=true`. The PR description and changelog now state that distinction.

CHANGE_LOG.md
bghcore: Updated the changelog to leave an empty `Unreleased` section and place these notes under dated `2.0.0`, matching the sibling release convention.

chat-widget/samples/typescript-sample/package.json
bghcore: Updated the TypeScript sample dependency to `@microsoft/omnichannel-chat-widget@^2.0.0` so it exercises the release produced by this PR.

chat-widget/scripts/validate-consumer-audit.cjs
bghcore: Follow-up from the exact packed-consumer CI tree: the prior URL-only gate had also reviewed `@babel/runtime-corejs3@7.20.13` under WebChat and a hoisted `uuid@8.3.2` shared by WebChat and chat SDK/Azure Signaling. The validator now records allowed dependency roots per exact package/version/advisory. Markdown/sanitize regressions remain WebChat-only, while only the known UUID 8.3.2 finding permits both reviewed roots.

coryeurom (approved):

ca1eb (approved):

PR #974, opened 2026-08-17
fix(chat-widget): prevent duplicate conversations from concurrent start chat

coryeurom:
## Summary

Fixes a race condition where two `prepareStartChat()` calls could fire concurrently if the chat button is clicked rapidly (within ~500ms), resulting in **two conversations created on the backend** — one of which is orphaned and never ended (no `ConversationEndedByCustomer` event).

[AB#6114594](https://github.com/microsoft/omnichannel-chat-widget/pull/974#issue-5173631493)
related to https://portal.microsofticm.com/imp/v5/incidents/details/21000001625459/summary

## Root Cause

- `ChatButtonStateful`'s `onClick` (`() => ref.current()`) does not await the result and has no debounce or "in-progress" guard.
- Every guard in the `prepareStartChat` → `setPreChatAndInitiateChat` → `initStartChat` chain checks `conversationState === Closed` using **React state**, which updates asynchronously. The button stays visible and clickable until React re-renders after the first call dispatches `SET_CONVERSATION_STATE: Loading`, so the second call still observes the stale `Closed` state and reaches `facadeChatSDK.startChat()`.
- Both calls operate on the same `facadeChatSDK` singleton (shared mutable `requestId`, and a stale `conversation` object surviving from the previous session).

## Changes

| File | Change |
|---|---|
| `livechatwidget/common/startChat.ts` | Module-level mutexes (`runExclusively`) wrapping `prepareStartChat` and `initStartChat`. A concurrent invocation is dropped and reported via telemetry. |
| `chatbuttonstateful/ChatButtonStateful.tsx` | `startChatInProgress` ref guard rejects rapid clicks optimistically, without waiting for a React re-render. |
| `livechatwidget/common/endChat.ts` | `chatSDKStateCleanUp` now also clears the SDK's `conversation` object (mirroring `OmnichannelChatSDK.endChat()`), and no-ops on a missing `chatSDK`. |
| `common/telemetry/TelemetryConstants.ts` | New `ConcurrentStartChatRejected` event, mapped to `ScenarioType.ACTIONS`. |

### Design notes

- **Two separate guards, not one.** `prepareStartChat` calls `initStartChat` internally; a single shared guard would make the nested call see its own parent as in-flight and drop it.
- **Concurrent callers return immediately** rather than awaiting the in-flight promise. This removes any possibility of a future re-entrant call deadlocking on itself, and no caller acts on the resolved value (verified across `ChatButtonStateful`, `ProactiveChatPaneStateful`, and the `BroadcastEvent.StartChat` subscriber, which is fire-and-forget).
- **Guard is released in a `finally`**, so a failed start chat still lets the user retry.
- **Defense in depth:** the button-level ref guard stops the common case before it reaches the SDK layer; the module-level mutexes also cover the `BroadcastEvent.StartChat` subscriber, `ProactiveChatPaneStateful`, `PreChatSurveyPaneStateful`, and `ReconnectChatPaneStateful` entry points.

Covers all four "Proposed Investigation Areas" from the issue. Verified `BroadcastEvent.StartChat` is never posted from within the widget's own start-chat flow, and that `handleStartChatError` / `handleChatReconnect` do not re-enter `initStartChat`, so no deadlock or dropped-nested-call path exists.

## Tests

New `startChatConcurrency.test.ts` (4 tests):
- Two concurrent `prepareStartChat` calls → `facadeChatSDK.startChat` invoked **once**, `ConcurrentStartChatRejected` logged
- Two concurrent `initStartChat` calls → invoked **once**
- Guard is released so a subsequent sequential start chat is allowed
- An in-flight failure is not surfaced to the dropped caller, and the guard is still released

New `ChatButtonStateful` tests (2):
- Rapid double-click → `startChat` called **once**
- A new click after the previous `startChat` settles is allowed

Each new test was verified to **fail** without the source fix (stashed the two source files and re-ran) and pass with it.

## Validation

- `yarn test:unit` — **60/60 suites, 713/713 tests pass**
- `npx tsc --noEmit -p tsconfig.json` — clean
- `npx eslint` on all changed files — clean
- `yarn build:esm` — succeeds

## Risk

Low. No public API surface changes. Behavior only differs when a second start-chat is attempted while one is already in flight — previously a duplicate backend conversation, now a no-op with telemetry.


bghcore (approved):
## Review (AB#6114594 / ICM 21000001625459)

Pulled the work item and the linked Restricted CRI (Sev 3, Active, LiveChat+C2). Recurrence of [749126979](https://portal.microsofticm.com/imp/v5/incidents/details/749126979/summary).

**What the incident actually says (this is stricter than the PR summary):**
- Two *valid* `startChat` requests leave **one browser session**. MessagingRuntime/ACD then correctly create and route one conversation each. Backend is not the bug.
- **Does not necessarily mean the customer clicked twice.** The prior case measured two client requests **559ms** apart. This CRI’s backend first-seen times are ~4.5s apart (processing lag, not proof of click interval).
- Customer is on **legacy LCW 1.8.4**. Discussion states newer-widget protection already shipped and this version does not have it. 6114594 is the named equivalent correction.
- Expected result: one CRM conversation per portal session.

The in-flight mutex + button ref + `conversation = null` cleanup is the right shape for the overlapping-call race the CRI/code inspection called out. Approving; the items below are non-blocking but should be fixed or answered before we treat this as the CRI close.

### Blocking for the CRI? No — but please address

**1. CHANGE_LOG contradicts the code (nit, please fix in this PR)**

`CHANGE_LOG.md` says a concurrent invocation **“awaits the in-flight start”**. `runExclusively` **drops** it and returns. Also tag the entry `Bug 6114594` / `AB#6114594` like neighboring rows.

**2. Spec gap: “not necessarily a double-click”**

Button-ref + `prepareStartChat` mutex covers the 559ms overlap and `BroadcastEvent.StartChat` / proactive / prechat / reconnect while a start is *in flight*. That matches the CRI’s overlapping-request story.

It does **not** cover a second start after the first `startChat` has settled and the mutex/`startChatInProgressRef` have been released, if `conversationState` is still `Closed` (failed start, prechat return, host calling `StartChat` again). The 4.5s backend gap is not enough to prove that path, but the CRI wording is “one session, two requests,” not “double-click only.” Worth a sentence in the PR (and ideally a test) that a `BroadcastEvent.StartChat` racing the button is dropped, and that a sequential retry after failure still works.

**3. Prepare/init cross-drop (edge, non-blocking for the button race)**

Two separate guards are required so `prepareStartChat → initStartChat` is not self-dropped. The flip side: if `initStartChat` is already in flight (prechat submit, reconnect, `hideStartChatButton` bootstrap), a later `prepareStartChat` can finish Loading / MaximizeChat and then have its nested `initStartChat` **dropped**. Unlikely for the 559ms button case; real if broadcast/bootstrap races prechat. Consider setting `guard.inFlight` *before* invoking `operation()` so a sync re-entry in the preamble cannot miss the lock.

**4. `chatSDKStateCleanUp` is incomplete vs SDK `endChat()`**

Nulling `conversation` matches `OmnichannelChatSDK.endChat()`’s `finally`. The SDK also `conversation?.disconnect()` first. If cleanup runs after a partial join, we can leave the ACS session open — same orphan class the CRI is about. No unit test that cleanup nulls `conversation` or no-ops on a missing SDK.

**5. Ship path for 1.8.4**

CRI customer is on **1.8.4**. Confirm this `main` change is the 1.8.4-line fix (vs a different “newer widget” already claimed shipped) and call out the patch/release plan so the CRI can get a real ETA.

### What looks good
- Defense in depth: button ref (optimistic) + module mutexes (all entry points) + SDK field reset.
- Drop-not-await is the right deadlock posture; callers are fire-and-forget.
- `ConcurrentStartChatRejected` is useful for proving the CRI is gone.
- Concurrency tests fail without the source change (as described). `yarn test:unit` 713/713 claimed.

**Verdict:** Approve. Core overlapping `startChat` race is fixed. Please land the CHANGE_LOG correction and a note on 1.8.4 ship + the non-click entry points.

MohammadSuraj (commented):
File Name - startChat.ts , line 53:

[Major] Scope the mutex to the widget instance
This lock is module-wide, while the component supports independent widget instances through  widgetInstanceId  and per-instance  FacadeChatSDK  objects. If two widgets legitimately start simultaneously, widget A sets this guard and widget B is treated as a duplicate and silently dropped.
Please key the in-flight operation by widget/cache ID (or  FacadeChatSDK ) and add a test proving two distinct widget instances can start concurrently.

coryeurom:
Good catch — fixed in a774e3a8.

The two locks are now `WeakMap`s keyed by the widget's `FacadeChatSDK` instance instead of being module-wide:

```ts
const prepareStartChatGuards = new WeakMap<object, IStartChatGuard>();
const initStartChatGuards = new WeakMap<object, IStartChatGuard>();

const getGuard = (guards: WeakMap<object, IStartChatGuard>, facadeChatSDK?: FacadeChatSDK): IStartChatGuard => {
    const owner: object = facadeChatSDK ?? unkeyedGuardOwner;
    ...
};
```

I keyed on `FacadeChatSDK` rather than the widget cache ID because:

- It's a required parameter on both `prepareStartChat` and `initStartChat`, whereas `props` is optional on `initStartChat`, so the cache ID isn't always derivable.
- `LiveChatWidget` creates exactly one per widget via `useState` (`LiveChatWidget.tsx:27,51`), so it's stable across re-renders and unique per instance — including the case where two widgets share an `orgId`/`widgetId` and no `widgetInstanceId` is set, which would collide on cache ID.
- `WeakMap` means no retention of torn-down widgets.

The rejection telemetry now also carries `CustomProperties: { WidgetCacheId }` so a dropped call can be attributed to a specific widget in Kusto. There's a fallback owner object for the degenerate case where no `FacadeChatSDK` is available, so the guard degrades to the previous behaviour rather than disappearing.

**Tests** — added a `multiple widget instances` block (`startChatConcurrency.test.ts`) with 4 tests:

1. two distinct instances can run `prepareStartChat` concurrently — both reach `startChat()`, no rejection telemetry
2. same for `initStartChat`
3. a duplicate call for instance A is *still* dropped while instance B is in flight (A→1, B→1, exactly one rejection, attributed to A)
4. each instance's guard is released independently

I verified tests 1–3 genuinely catch this: temporarily forcing `getGuard` back to a single shared owner makes all three fail, and reverting makes them pass.

Suite is green: 60/60 suites, 718/718 tests; `tsc --noEmit` and `eslint` clean; `build:esm` succeeds.

Also corrected the changelog entry, which was stale on two counts — it described the guards as "module-level" and said concurrent invocations "await the in-flight start" (they're dropped immediately, to avoid any possibility of a re-entrant call deadlocking on itself).

One thing I deliberately left alone: the module-level `let widgetInstanceId` / `let optionalParams` at the top of `startChat.ts` (lines 30–34) are also shared across widget instances and `prepareStartChatInternal` overwrites them on every call. That's pre-existing and orthogonal to the locking, so I kept it out of this PR — happy to file it separately if you'd like.

chat-widget/src/components/livechatwidget/common/startChat.ts
MohammadSuraj: [Major] Keep all start-chat state isolated per widget

The new guards are correctly scoped per  FacadeChatSDK , but  optionalParams  and  widgetInstanceId  remain module-level mutable variables. Two different widgets can now execute concurrently and overwrite each other’s state.

I reproduced this with two widget instances using different  customContext  values: widget A’s  startChat  received widget B’s context.

Please make these values local/per-widget rather than module-global, and add a test confirming concurrent widgets retain their own custom context.

deeptijoshim (approved):

chat-widget/src/components/livechatwidget/common/startChat.ts
coryeurom: Good catch, and thanks for taking the time to actually reproduce it — you're right, and it's worse than "leftover globals": my per-instance guard change is what made this reachable. The earlier module-wide lock happened to serialize every widget on the page, so the shared state was never contended. Scoping the lock per `FacadeChatSDK` removed that accidental serialization and exposed the real race.

Fixed in 53fc9578.

**The window**

`initStartChatInternal` did `await setCustomContextParams(state, props)` and then read the module-level `optionalParams` *after* that await resolved. Everything before the await in the other widget's `initStartChatInternal` is synchronous, so the interleaving is deterministic:

1. Widget A runs `setCustomContextParams` → sets `optionalParams = { customContext: A }`, returns, suspends on the `await`.
2. Widget B runs its synchronous prologue and its own `setCustomContextParams` → overwrites `optionalParams = { customContext: B }`.
3. Widget A resumes and reads `optionalParams` → gets B's context.

Which is exactly what you saw.

**Change**

- `setCustomContextParams` now returns the resolved `StartChatOptionalParams` instead of mutating shared state; the caller keeps it in a local and passes it straight into `Object.assign`.
- The widget cache id used to look up persisted custom context is now computed locally from `props` instead of being read from a module-level variable seeded by `prepareStartChatInternal`. Same value — both derived from the same `props` — just no longer shared. It also removes the `getWidgetCacheIdfromProps(undefined)` throw that was previously only avoided because `prepareStartChat` happened to seed the variable first.
- `popoutWidgetInstanceId` is now a local `const` in `canStartPopoutChat`.
- Removed the now-dead resets in `prepareStartChatInternal` and the `initStartChatInternal` `finally` block.

No module-level mutable state remains in the file.

**Tests**

Added two tests under `custom context isolation across concurrent widget instances`, covering both resolution paths — custom context from widget state, and custom context resolved from cache by widget cache id:

```
√ should start each widget with the custom context from its own state
√ should start each widget with the persisted custom context for its own widget cache id
```

I verified these actually catch the bug rather than just passing: I temporarily reinstated the module-level variable and re-ran the suite — both new tests fail (widget A receives widget B's context), and both pass once reverted. The other 9 concurrency tests pass in both states, which is why this needed dedicated coverage.

Full suite: 60 suites / 720 tests passing. `tsc --noEmit`, `eslint`, and `build:esm` clean.

MohammadSuraj (approved):

bghcore:
## Follow-up review (`7dd419f`)

Re-checked HEAD against the earlier comments (mine + Suraj). Approvals already on the thread: me on `a393d7bf`, Deepti on `7d5b362`, Suraj on HEAD after the custom-context fix.

### Addressed

- **Suraj — module-wide mutex.** Fixed in `a774e3a8`. Locks are `WeakMap`s keyed by `FacadeChatSDK`. Multi-instance tests cover concurrent prepare/init, same-instance drop while another instance is in flight, and independent release.
- **Suraj — `optionalParams` / `widgetInstanceId` leak.** Fixed in `53fc9578`. `setCustomContextParams` returns a local object; cache id is computed from `props`. Isolation tests cover state and cache paths. No module-level mutable start-chat state left.
- **CHANGE_LOG “awaits the in-flight start”.** Wording is now drop-not-await. Correct.

### Still open from the earlier review

These were non-blocking for the overlap race, but they are still not done. I would not treat this as CRI-close ready until 1 and 5 are answered.

1. **CHANGE_LOG still has no work-item tag.** Neighboring rows use `Bug 6525143` / `AB#6523665`. This entry is still just `[Bug]`. Please add `AB#6114594`.

2. **Spec gap: not necessarily a double-click — only partial.** The code comment now mentions a `BroadcastEvent.StartChat` race, and that subscriber does call `prepareStartChat`, so the mutex covers the overlap. Failure-then-retry is tested. Still missing: a dedicated broadcast-vs-button test, and the PR body still frames the bug as “clicked rapidly (~500ms)”.

3. **Prepare/init cross-drop — not done.** `runExclusively` is still:

   ```ts
   const inFlight = operation();
   guard.inFlight = inFlight;
   ```

   `inFlight` is assigned *after* `operation()` starts, so a sync re-entry in the preamble can miss the lock. Two maps are still required so `prepare → init` is not self-dropped; the flip side is unchanged: if `initStartChat` is already in flight (prechat submit, reconnect, `hideStartChatButton` bootstrap), a later `prepareStartChat` can Loading/Maximize and then have its nested `initStartChat` dropped.

4. **`chatSDKStateCleanUp` is still incomplete vs SDK `endChat()`.** Still only `conversation = null`. No `disconnect()`. No unit test — there is no `endChat.test.ts`, and the start-chat suites mock the helper.

5. **1.8.4 ship path / CRI ETA — not done, and more important now.** CRI customer is on **1.8.4**. This PR is landing on `main` after the 2.0.0 prep (`#975`). This is not the 1.8.4 line. Need an explicit patch/hotfix plan so the CRI can get a real ETA.

### New on this pass

- Both new `[Bug]` rows sit under `## [2.0.0]` with an empty `[Unreleased]`. This PR is still open. If 2.0.0 is already cut, this attributes an unshipped CRI fix to a released version. Move both rows to `[Unreleased]` and tag `AB#6114594`.
- PR summary still says “module-level mutexes”. Stale after `a774e3a8`.

### Verdict

Overlapping `startChat` race, per-widget locking, and the custom-context leak are fixed. **Do not treat this as CRI-close ready** until the 1.8.4 ship path is written down and the changelog placement / work-item tag are corrected. #3 and #4 are still the same non-blocking holes.

PR #982, opened 2026-09-03
fix(chat): validate cached Loading conversations before new chat

MSSunnyNayak:
## Problem

During a page reload or full-page navigation, the widget can restore a cached conversation in the transient `Loading` state. The existing recovery gate accepts cached context only when the local state is exactly `Active`, so it can skip authoritative conversation validation and proceed toward creating a new chat.

## Root cause

Recovery eligibility is decided from local UI state before `checkIfConversationStillValid()` consults the service. A cached conversation with stable identity but a transient `Loading` state is rejected before that validation call.

## Fix

| Change | Details |
|--------|---------|
| **Recovery eligibility** | Treat cached `Active` and `Loading` states as validation candidates when a nonempty request ID is available. |
| **Reconnect precedence** | Preserve the existing `reconnectId` path as the highest-priority recovery mechanism. |
| **Safety exclusions** | Continue to reject terminal, post-chat, proactive, out-of-office, reconnect-choice, malformed, and missing-identity states. |
| **Telemetry** | Add `RecoveryEligibilityEvaluated` with only allowlisted state/reason strings and booleans; no IDs, tokens, context payloads, or customer content. |
| **Scope** | Change only the initialization path that immediately performs authoritative conversation validation. Alternate paths without equivalent validation remain unchanged. |

## Pipeline security-audit refresh

The first PR run was interrupted by `sanitize-html` advisories published after the previous successful build. The follow-up commit:

- pins the widget's direct `sanitize-html` dependency to patched version `2.17.7`;
- records the two newly published advisories that remain only under the required `botframework-webchat` hotfix's exact nested `sanitize-html@2.14.0` dependency;
- keeps audit exceptions scoped to that exact version, advisory, and dependency owner.

## Testing

- 38 recovery eligibility and initialization tests passed.
- 16 related `startChat` tests passed.
- Changed-file ESLint passed with zero warnings.
- The packed-consumer dependency tree contains direct `sanitize-html@2.17.7` and nested `sanitize-html@2.14.0` only under `botframework-webchat`.
- The expected nested advisory ownership set passes `validate-consumer-audit.cjs`.
- Three independent correctness, architecture, and telemetry/privacy reviews found no blocking issues.

## Follow-up

This is a contained recovery-gate fix. A durable follow-up should represent validation as `Valid`, `Terminal`, or `Unknown`, and must not allow a transient validation failure to silently authorize new-chat creation.

Fixes AB#6667146

### DIFF
--- a/CHANGE_LOG.md
+++ b/CHANGE_LOG.md
@@ -16,6 +16,9 @@ All notable changes to this project will be documented in this file.
 - Stopped Storybook deploy from running on `w-v*` tags. Pushes to `main` still deploy Storybook.
 - Consolidated the Chat Widget 2.0.0 notes into one Breaking, Changed, Added, Tests, Fixed, and Security section and removed mid-auth entries.
 
+### Fixed
+- [Bug 6667146] Cached chat recovery now validates a persisted `Loading` conversation when a request ID is still available, instead of treating the transient UI state as a new-chat signal. Recovery eligibility emits privacy-safe reason-code telemetry and continues to exclude explicit terminal, post-chat, proactive, out-of-office, and reconnect-choice states.
+
 ## [2.0.0] - 2026-08-18
 
 ### Breaking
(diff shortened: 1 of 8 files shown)


MSSunnyNayak:
Pipeline diagnosis: the recovery change passed unit tests, sample build, and CommonJS validation. The failed step was the packed-consumer audit after new sanitize-html advisories were published. Commit 40a6357f pins the direct dependency to 2.17.7 and scopes the unavoidable 2.14.0 findings to the required WebChat hotfix dependency. A replacement run was triggered automatically.

MSSunnyNayak:
Telemetry-first follow-up is now available in #984. This behavioral proposal remains draft and is deferred while #984 collects the exact recovery guard outcome in production; it has not been closed so the reviewed containment is preserved.

PR #983, opened 2026-09-10
fix(security): refresh sanitize-html audit baseline

MSSunnyNayak:
## Problem

New `sanitize-html` advisories published after the previous successful build caused the packed-consumer audit on unrelated PRs to fail.

## Fix

- Pin the widget's direct runtime dependency to patched `sanitize-html@2.17.7`.
- Record the two new advisories that remain only under the required `botframework-webchat` hotfix's exact `sanitize-html@2.14.0` dependency.
- Keep audit exceptions constrained by exact package version, advisory URL, and dependency owner.

## Validation

- Packed consumer resolves direct `sanitize-html@2.17.7`.
- Nested `sanitize-html@2.14.0` is owned only by the required WebChat hotfix.
- The expected nested advisory set passes `validate-consumer-audit.cjs`.
- Changed-file lint passes.

This prerequisite keeps feature PRs from mixing security-audit maintenance with their functional changes.


MSSunnyNayak:
Telemetry-only PR #984 is stacked on this security-audit prerequisite so the diagnostic commit remains isolated from audit maintenance.

bghcore (approved):

SriDeviMallipudi (approved):

